PR #10, opened 2020-09-14
Added SSLVPN connection count per VDOM and IPSec status/traffic gauge

Boolman:
Thanks for this project!

sslvpn connection count
ipsec tunnel status + rx/tx traffic
I also changed the Dockerfile, greatly reducing the image size. 
For some reason the interface speed was float so I changed the struct, not sure if this is differ from version to version

I just now realize I pushed some tls tweaking I had to to due to our slow firewall, might not be needed.


forti_token_client.go
```diff
@@ -38,6 +40,8 @@ type fortiTokenClient struct {
 }
 
 func (c *fortiTokenClient) newGetRequest(url string) (*http.Request, error) {
+	http.DefaultTransport.(*http.Transport).TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
```
bluecmd, line 43: Can you make this a flag setting?
In my environment all firewalls have correct certificates, so running this is a security issue. However, I can certainly see what some folks would like to be able to skip this check.

forti_token_client.go
```diff
@@ -38,6 +40,8 @@ type fortiTokenClient struct {
...
+	http.DefaultTransport.(*http.Transport).TLSHandshakeTimeout = 30 * time.Second
```
bluecmd, line 44: Same here, probably better to make this a flag to avoid magical constants. Something like --https-timeout=30s might be cool.

probe.go
```diff
@@ -191,6 +191,107 @@ func probeSystemVDOMResources(c FortiHTTP, registry *prometheus.Registry) bool {
...
+	}
+
+	return true
+
+
+}
+func probeIPSec(c FortiHTTP, registry *prometheus.Registry) bool {
+	var (
+		status = prometheus.NewGaugeVec(
+			prometheus.GaugeOpts{ // change type?
```
bluecmd, line 227: This should be a gauge as far as I can tell, so the comment might be ok to remove?

probe.go
```diff
@@ -191,6 +191,107 @@ func probeSystemVDOMResources(c FortiHTTP, registry *prometheus.Registry) bool {
...
+
+	)
+
+	registry.MustRegister(status)
+	registry.MustRegister(transmitted)
+	registry.MustRegister(received)
+
+	type proxyid struct {
+		Name    string `json:"p2name"`
+		Status    string `json:"status"`
```
bluecmd, line 255: Please run `go fmt`.

probe.go
```diff
@@ -191,6 +191,107 @@ func probeSystemVDOMResources(c FortiHTTP, registry *prometheus.Registry) bool {
...
+		VDOM    string
+	}
+	var res []ipsecResult
+	if err := c.Get("api/v2/monitor/vpn/ipsec", "vdom=*", &res); err != nil {
+		log.Printf("Error: %v", err)
+		return false
+	}
+	for _, v := range res {
+		for _, i := range v.Results {
+			if i.Type == "dialup" {
```
bluecmd, line 275: Can you add a comment why these are skipped?

testdata/ipsec.jsonnet
```diff
@@ -0,0 +1,51 @@
...
+        "incoming_bytes": 14298240,
+        "outgoing_bytes": 14248560,
+        "rgwy":"1.2.3.4"
+      }
+    ],
+    "vdom":"root",
+    "path":"vpn",
+    "name":"ipsec",
+    "status":"success",
+    "serial":"FGVMEVZFNTS3OAC8",
```
bluecmd, line 47: Consider if you want to censor your serial number - it seems to be an EVAL though so maybe alright.

probe.go
```diff
@@ -409,7 +510,7 @@ func probeInterfaces(c FortiHTTP, registry *prometheus.Registry) bool {
 		Name      string
 		Alias     string
 		Link      bool
-		Speed     int
+		Speed     float64
```
bluecmd, line 513: Could you modify one of the interfaces in the test file (`testdata/interface.jsonnet`) to return a float number and add a comment in the test file that some versions appear to use floats?

bluecmd (changes requested):
Good stuff, thanks!

Comments about `go fmt` as well as making some options flags.

bluecmd:
Oh! And please update the readme.md file to include the new metrics, thanks!

Dockerfile
```diff
@@ -1,11 +1,17 @@
...
 COPY . .
 RUN go get -v -t -d ./...
+RUN CGO_ENABLED=0 go build -o main .
+
+FROM scratch
+WORKDIR /opt/fortigate_exporter
 
-RUN go build -o main .
+COPY --from=builder /build/main .
+COPY ca-certificates.crt .
```
bluecmd, line 13: I have been thinking about ca-certificates.crt. That seems like something a bit iffy for this repo to carry - it would quickly be out of date, and I not really a file you'd want to download from a random github repo I suppose. Is it possible to include Either https://curl.haxx.se/ca/cacert.pem or copy ca-certificates from the "builder"? I'm a bit of a Docker rookie.

Boolman:
yes sure, 
`COPY --from=builder /etc/ssl/certs/ca-certificates.crt .`

this should work just fine

bluecmd (approved):

bluecmd:
Looks good to me, @Boolman are you happy with this or do you have more changes you want in before I merge this?

Boolman:
No im all gooood, thanks.

Issue #12, opened 2020-10-07
move fortigate-key.yaml out of /opt to allow Kubernetes Secret mounts


kifeo:
Kubernetes only support mounting of full directories, not specific files for Secrets mounts.

Mounting the whole /opt as read only for Secrets makes the container not start.

Please change it to some directory not used by the system, like /config so this does not interfere with the mount.
Thanks

The kubelet error is here for reference: 
`  failed to start container "fortigate-exporter": Error response from daemon: OCI runtime create failed: container_linux.go:349: starting container process caused "process_linux.go:449: container init caused \"mkdir /opt/fortigate_exporter: read-only file system\"": unknown`

bluecmd:
Sounds good to me! Should be fixed now with your PR (thanks!)

PR #13, opened 2020-10-07
changed fortigate-key.yaml to /config

kifeo:
Changed fortigate-key.yaml to /config instead of /opt so this can be Kubernetes Secret mounted

referenced in https://github.com/bluecmd/fortigate_exporter/issues/12


bluecmd (approved):

bluecmd:
Thanks!

PR #18, opened 2020-10-16
added testdata/fsw-interface.jsonnet

kifeo:


testdata/fsw-interface.jsonnet
```diff
@@ -0,0 +1,880 @@
+{
```
bluecmd, line 1: Can you add a comment for what URL this was fetched from? That makes it easy for somebody else to update this.

Thanks!

bluecmd (changes requested):

testdata/fsw-interface.jsonnet
kifeo: I added it, thanks

bluecmd (approved):

bluecmd:
Thanks!

PR #27, opened 2021-01-04
retype metrics

### LINKED ISSUE #5: Retype some metrics to counters
From @bbrazil:

> fortigate_policy_hit_count_total is a counter so should have type counter, and similarly with the two below it. The _count there also seems redundant, I'd suggest just hits.

That makes sense, I don't know why I made them gauges in the first place - seems to be counters from taking a look now.

bluecmd:
This fixes #5. Some metrics have accidentally been set to gauges,
these are now counters.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ Per-VDOM:
  * `fortigate_interface_receive_bytes_total`
  * `fortigate_interface_transmit_errors_total`
  * `fortigate_interface_receive_errors_total`
- * `fortigate_vpn_connections_count_total`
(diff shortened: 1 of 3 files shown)


bluecmd:
@Boolman I noticed that the `fortigate_vpn_connections_count_total` is really a gauge as it measures the value going up/down - and thus `_total` shouldn't be part of the suffix. `_count` could also be considered redundant, so I chose to remove it as well.

This will of course break metric naming, but I think it's worth doing.

PR #29, opened 2021-01-08
Add License metrics

kcbark:
Metrics to monitor vdom license usage. Should be possible to add other license statuses as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ Global:
  * `fortigate_cpu_usage_ratio`
  * `fortigate_memory_usage_ratio`
  * `fortigate_current_sessions`
- * `fortigate_license_status`
+ * `fortigate_license_vdom_info`
 
 Per-VDOM:
 
--- a/README.md
+++ b/README.md
@@ -18,7 +18,8 @@ Global:
  * `fortigate_cpu_usage_ratio`
  * `fortigate_memory_usage_ratio`
  * `fortigate_current_sessions`
- * `fortigate_license_vdom_info`
+ * `fortigate_license_vdom_usage`
+ * `fortigate_license_vdom_max`
 
 Per-VDOM:
 
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@ Global:
  * `fortigate_cpu_usage_ratio`
  * `fortigate_memory_usage_ratio`
  * `fortigate_current_sessions`
+ * `fortigate_license_status`
 
 Per-VDOM:
 
--- a/probe.go
+++ b/probe.go
@@ -34,10 +34,12 @@ type ProbeCollector struct {
 type probeFunc func(FortiHTTP) ([]prometheus.Metric, bool)
 
 func probeSystemStatus(c FortiHTTP) ([]prometheus.Metric, bool) {
-	mVersion := prometheus.NewDesc(
-		"fortigate_version_info",
-		"System version and build information",
-		[]string{"serial", "version", "build"}, nil,
+	var (
+		mVersion = prometheus.NewDesc(
+			"fortigate_version_info",
+			"System version and build information",
+			[]string{"serial", "version", "build"}, nil,
+		)
 	)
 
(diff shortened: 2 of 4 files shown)


probe.go
```diff
@@ -630,6 +629,50 @@ func probeHAStatistics(c FortiHTTP) ([]prometheus.Metric, bool) {
...
+	var (
+		vdomInfo = prometheus.NewDesc(
+			"fortigate_license_vdom_info",
+			"Info metric regarding vdom licenses",
+			[]string{"type"}, nil,
+		)
+	)
+
+	type LicenseStatus struct {
+		Vdom struct {
```
bluecmd, line 642: Nit: s/Vdom/VDOM (as per https://github.com/golang/go/wiki/CodeReviewComments#initialisms)

probe.go
```diff
@@ -630,6 +629,50 @@ func probeHAStatistics(c FortiHTTP) ([]prometheus.Metric, bool) {
...
+			Used       int    `json:"used"`
+			Max        int    `json:"max"`
+		} `json:"vdom"`
+	}
+
+	type LicenseResponse struct {
+		HTTPMethod string        `json:"http_method"`
+		Results    LicenseStatus `json:"results"`
+		VDOM       string        `json:"vdom"`
+		Path       string        `json:"path"`
```
bluecmd, line 654: You can omit the fields you are not using if you wish. Up to you :-)

README.md
```diff
@@ -18,6 +18,7 @@ Global:
  * `fortigate_cpu_usage_ratio`
  * `fortigate_memory_usage_ratio`
  * `fortigate_current_sessions`
+ * `fortigate_license_vdom_info`
```
bluecmd, line 21: What do you think about having `fortigate_license_vdom_usage` and `fortigate_license_vdom_max` instead?
My thinking is that it makes it easier to discover, and currently it kind of breaks the [following](https://prometheus.io/docs/practices/naming/#metric-names) best-practice section:

> As a rule of thumb, either the sum() or the avg() over all dimensions of a given metric should be meaningful (though not necessarily useful). If it is not meaningful, split the data up into multiple metrics. For example, having the capacity of various queues in one metric is good, while mixing the capacity of a queue with the current number of elements in the queue is not.

Doing a `sum(fortigate_license_vdom_info)` doesn't make much sense as it contains both `max` and `usage`

testdata/license-status.jsonnet
```diff
@@ -0,0 +1,25 @@
+{
```
bluecmd, line 1: Please add a comment in the first row of what URL one can use to collect the API request from. See the other jsonnet files for example.

bluecmd (commented):
Cool! Thanks for your contribution!
I have some comments.

README.md
kcbark: Yeah, makes more sense. 👍

probe.go
```diff
@@ -630,6 +629,47 @@ func probeHAStatistics(c FortiHTTP) ([]prometheus.Metric, bool) {
 	return m, true
 }
 
+func probeLicenseStatus(c FortiHTTP) ([]prometheus.Metric, bool) {
+	var (
+		vdomUsed = prometheus.NewDesc(
+			"fortigate_license_vdom_usage",
+			"Info metric regarding used vdom licenses",
```
bluecmd, line 636: Update the description of these metrics. Maybe something like "The amount of VDOM licenses currently used"  and "The total amount of VDOMs licenses available"

bluecmd (changes requested):

bluecmd:
Looks good, only the description left I'd say :-)

probe.go
kcbark: Yeah, that's a better wording, thanks :)

kcbark:
All good? :)

bluecmd (approved):

bluecmd:
🎉 Thanks!

PR #33, opened 2021-02-15
feat: implement link-monitor API

secustor:
Implements the link-monitor API. 

Is needed here to get the jitter of WAN links

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,18 @@ Per-VDOM:
  * `fortigate_ha_member_bytes_total`
  * `fortigate_ha_member_ips_events_total`
 
+ Per-Link and VDOM: 
+ * `fortigate_link_status`
+ * `fortigate_link_latency_seconds`
+ * `fortigate_link_latency_jitter_seconds`
+ * `fortigate_link_packet_loss_ratio`
+ * `fortigate_link_packet_sent_total`
+ * `fortigate_link_packet_received_total`
+ * `fortigate_link_active_sessions`
+ * `fortigate_link_bandwidth_tx_byte_per_second`
+ * `fortigate_link_bandwidth_rx_byte_per_second`
+ * `fortigate_link_status_change_time_seconds`
+
 ## Usage
 
 Example:
--- a/probe.go
+++ b/probe.go
@@ -670,6 +670,121 @@ func probeLicenseStatus(c FortiHTTP) ([]prometheus.Metric, bool) {
 	return m, true
 }
 
+func probeLinkMonitor(c FortiHTTP) ([]prometheus.Metric, bool) {
+	var (
+		linkStatus = prometheus.NewDesc(
+			"fortigate_link_status",
+			"Signals if the status is up (1) or down (0)",
+			[]string{"vdom", "monitor", "link"}, nil,
+		)
+		linkLatency = prometheus.NewDesc(
+			"fortigate_link_latency_seconds",
+			"Average latency of this link based on the probe interval set in the monitor in seconds",
+			[]string{"vdom", "monitor", "link"}, nil,
+		)
+		linkJitter = prometheus.NewDesc(
+			"fortigate_link_latency_jitter_seconds",
+			"Average of the latency jitter  on this link based on the probe interval set in the monitor in seconds",
+			[]string{"vdom", "monitor", "link"}, nil,
+		)
+		linkPacketLoss = prometheus.NewDesc(
+			"fortigate_link_packet_loss_ratio",
+			"Percentage of packages lost relative all sent",
+			[]string{"vdom", "monitor", "link"}, nil,
+		)
+		linkPacketSent = prometheus.NewDesc(
+			"fortigate_link_packet_sent_total",
+			"Number of packets sent on this link",
+			[]string{"vdom", "monitor", "link"}, nil,
+		)
+		linkPacketReceived = prometheus.NewDesc(
+			"fortigate_link_packet_received_total",
+			"Number of packets received on this link",
+			[]string{"vdom", "monitor", "link"}, nil,
+		)
+		linkSessions = prometheus.NewDesc(
+			"fortigate_link_active_sessions",
+			"Number of sessions active on this link",
+			[]string{"vdom", "monitor", "link"}, nil,
+		)
+		linkBandwidthTx = prometheus.NewDesc(
+			"fortigate_link_bandwidth_tx_byte_per_second",
+			"Bandwidth available on this link for sending",
+			[]string{"vdom", "monitor", "link"}, nil,
+		)
+		linkBandwidthRx = prometheus.NewDesc(
+			"fortigate_link_bandwidth_rx_byte_per_second",
+			"Bandwidth available on this link for sending",
+			[]string{"vdom", "monitor", "link"}, nil,
+		)
+		linkStatusChanged = prometheus.NewDesc(
+			"fortigate_link_status_change_time_seconds",
+			"Unix timestamp describing the time when the last status change has occurred",
+			[]string{"vdom", "monitor", "link"}, nil,
+		)
+	)
+
+	type LinkMonitor struct {
+		Status         string  `json:"status"`
+		Latency        float64 `json:"latency"`
+		Jitter         float64 `json:"jitter"`
+		PacketLoss     float64 `json:"packet_loss"`
+		PacketSent     float64 `json:"packet_sent"`
+		PacketReceived float64 `json:"packet_received"`
+		Session        float64 `json:"session"`
+		TxBandwidth    float64 `json:"tx_bandwidth"`
+		RxBandwidth    float64 `json:"rx_bandwidth"`
+		StateChanged   float64 `json:"state_changed"`
+	}
+
+	type LinkGroup map[string]LinkMonitor
+
+	type linkMonitorResponse struct {
+		HTTPMethod string               `json:"http_method"`
+		Results    map[string]LinkGroup `json:"results"`
+		Vdom       string               `json:"vdom"`
+		Path       string               `json:"path"`
+		Name       string               `json:"name"`
+		Status     string               `json:"status"`
+		Serial     string               `json:"serial"`
+		Version    string               `json:"version"`
+		Build      int                  `json:"build"`
+	}
+
+	var r linkMonitorResponse
+
+	if err := c.Get("api/v2/monitor/system/link-monitor", "", &r); err != nil {
+		log.Printf("Error: %v", err)
+		return nil, false
+	}
+
+	m := []prometheus.Metric{}
+
+	for linkGroupName, linkGroup := range r.Results {
+		for linkName, link := range linkGroup {
+			// prepare values
+			var wanStatus float64
+			if link.Status == "up" {
+				wanStatus = 1
+			} else {
+				wanStatus = 0
+			}
+
+			m = append(m, prometheus.MustNewConstMetric(linkStatus, prometheus.GaugeValue, wanStatus, r.Vdom, linkGroupName, linkName))
+			m = append(m, prometheus.MustNewConstMetric(linkLatency, prometheus.GaugeValue, link.Latency/1000, r.Vdom, linkGroupName, linkName))
+			m = append(m, prometheus.MustNewConstMetric(linkJitter, prometheus.GaugeValue, link.Jitter/1000, r.Vdom, linkGroupName, linkName))
+			m = append(m, prometheus.MustNewConstMetric(linkPacketLoss, prometheus.GaugeValue, link.PacketLoss/100, r.Vdom, linkGroupName, linkName))
+			m = append(m, prometheus.MustNewConstMetric(linkPacketSent, prometheus.CounterValue, link.PacketSent, r.Vdom, linkGroupName, linkName))
+			m = append(m, prometheus.MustNewConstMetric(linkPacketReceived, prometheus.CounterValue, link.PacketReceived, r.Vdom, linkGroupName, linkName))
+			m = append(m, prometheus.MustNewConstMetric(linkSessions, prometheus.GaugeValue, link.Session, r.Vdom, linkGroupName, linkName))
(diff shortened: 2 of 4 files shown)


secustor:
@bluecmd

probe_test.go
```diff
@@ -626,3 +626,49 @@ func TestLicenseStatus(t *testing.T) {
...
+func TestLinkStatus(t *testing.T) {
+	c := newFakeClient()
+	c.prepare("api/v2/monitor/system/link-monitor", "testdata/link-monitor.jsonnet")
+	r := prometheus.NewPedanticRegistry()
+	if !testProbe(probeLinkMonitor, c, r) {
+		t.Errorf("probeLinkMonitor() returned non-success")
+	}
+
+	em := `
+# HELP fortigate_link_bandwidth_rx Bandwidth available on this link for sending
```
bluecmd, line 639: Add the unit as a suffix, like _bps for bit/s. I would also add "available".
So something like `fortigate_link_available_rx_bps` might be better? I am fine with `fortigate_link_bandwidth_rx_bps` as well if you prefer it the way it is.

probe_test.go
```diff
@@ -626,3 +626,49 @@ func TestLicenseStatus(t *testing.T) {
...
+# TYPE fortigate_link_bandwidth_rx gauge
+fortigate_link_bandwidth_rx{link="wan1",link_group="wan-mon",vdom="root"} 33557
+# HELP fortigate_link_bandwidth_tx Bandwidth available on this link for sending
+# TYPE fortigate_link_bandwidth_tx gauge
+fortigate_link_bandwidth_tx{link="wan1",link_group="wan-mon",vdom="root"} 68657
+# HELP fortigate_link_jitter electronical jitter on this link
+# TYPE fortigate_link_jitter gauge
+fortigate_link_jitter{link="wan1",link_group="wan-mon",vdom="root"} 1.1268666982650757
```
bluecmd, line 647: Unit is missing here as well, not clear what this is - hopefully the API defines it so it is not totally unknown?

Nit: Capital letter on the first letter of the description

probe_test.go
```diff
@@ -626,3 +626,49 @@ func TestLicenseStatus(t *testing.T) {
...
+fortigate_link_packet_loss_total{link="wan1",link_group="wan-mon",vdom="root"} 0
+# HELP fortigate_link_packet_received_total Number of packets received on this link
+# TYPE fortigate_link_packet_received_total counter
+fortigate_link_packet_received_total{link="wan1",link_group="wan-mon",vdom="root"} 278807
+# HELP fortigate_link_packet_sent_total Number of packets sent on this link
+# TYPE fortigate_link_packet_sent_total counter
+fortigate_link_packet_sent_total{link="wan1",link_group="wan-mon",vdom="root"} 278878
+# HELP fortigate_link_sessions Number of sessions active on this link
+# TYPE fortigate_link_sessions gauge
+fortigate_link_sessions{link="wan1",link_group="wan-mon",vdom="root"} 77
```
bluecmd, line 662: _active_sessions maybe would be better to signal that it is the current value as opposed to some cumulative count.

bluecmd (commented):
Very cool. Some metric naming suggestions.

probe_test.go
secustor: as the metric is called bandwidth in the backend, I prefer the second one.     
Like all other metrics in this API endpoint there is no documentation what these values mean.

probe_test.go
secustor: see comment above

probe_test.go
secustor: good idea

secustor:
I have found this two tutorials: 
https://docs.fortinet.com/document/fortigate/6.2.0/cookbook/382233/performance-sla-sla-targets    
https://docs.fortinet.com/document/fortigate/6.2.0/cookbook/478384/performance-sla-link-monitoring

These suggest that: 
- latency  is in milliseconds
- jitter is in milliseconds
- packet loss is in percentage

bluecmd:
Nice!

In order to help you out, because I think this is a great feature, I configured a link-monitor on a lab fortigate as well:
```
config system link-monitor
    edit "0"
        set srcintf "port3"
        set server "8.8.8.8"
        set update-cascade-interface disable
        set update-static-route disable
    next
end
```

This is the result of the API:
```json
    "results":{
      "0":{
        "port3":{
          "status":"up",
          "latency":1.0600999593734741,
          "jitter":0.049300007522106171,
          "packet_loss":0,
          "packet_sent":113,
          "packet_received":113,
          "session":55,
          "tx_bandwidth":43799,
          "rx_bandwidth":759780,
          "state_changed":1613408861
        }
      }
    },
```

I found this diag command described [here](https://www.linkedin.com/pulse/how-to-advanced-link-monitoring-fortigate-mohammedshahzan-mulla/):
```
ftg01 (bluecmd) #  diag sys link-monitor status

Link Monitor: 0, Status: alive, Server num(1), Flags=0x1 init, Create time: Mon Feb 15 18:07:41 2021
Source interface: port3 (9)
Interval: 500 ms
  Peer: 8.8.8.8(8.8.8.8)
        protocol: ping, state: alive
		Latency(Min/Max/Avg): 0.954/1.132/1.059 ms
		Jitter(Min/Max/Avg): 0.002/0.142/0.051
		Packet lost: 0.000%
		Number of out-of-sequence packets: 0
              	Fail Times(0/5)
              	Packet sent: 112, received: 112, Sequence(sent/rcvd/exp): 113/113/114
```

So it seems the jitter and latency are average, and the jitter might simply be the latency jitter (maybe stddev?).
I do not think it has to do with electrical jitter, and I do not believe it is a percentage given how it is displayed here.

A bit later I found this command as well:
```
ftg01 (bluecmd) #  diag sys link-monitor interface port3
Interface(port3): state(up, since Mon Feb 15 18:07:41 2021
), bandwidth(up:43952bps, down:748417bps), session count(49), tx(299608371876 bytes), rx(121607373822 bytes), latency(1.06), jitter(0.05), packet-loss(0.00).
```

So the up and down are definitely bps.

In summary, I think we should change the jitter to be `fortigate_link_latency_jitter_seconds` so it is clear it is latency jitter and not electrical jitter.

I will see if I can verify the packet_loss one as well by dropping some packets in the lab.

bluecmd:
Yep, it is percentage as you said:

On a Linux host I introduce a 5% packet loss, 10ms delay and 20ms latency jitter:
```
[debian]$ sudo tc qdisc change dev ens192 root netem delay 10ms 20ms loss 5%
```

On the Fortigate:
```
config system link-monitor
    edit "1"
        set srcintf "vlan2010"
        set server "172.16.31.10"
        set update-cascade-interface disable
        set update-static-route disable
    next
end
```

```
ftg01 (bluecmd) # diag sys link-monitor status
[..]
Link Monitor: 1, Status: alive, Server num(1), Flags=0x1 init, Create time: Mon Feb 15 18:19:47 2021
Source interface: vlan2010 (29)
Interval: 500 ms
  Peer: 172.16.31.10(172.16.31.10)
        Source IP(172.16.31.1)
        Route: 172.16.31.1->172.16.31.10/32, gwy(172.16.31.1)
        protocol: ping, state: alive
		Latency(Min/Max/Avg): 0.272/30.406/14.693 ms
		Jitter(Min/Max/Avg): 0.140/24.185/9.792
		Packet lost: 3.000%
		Number of out-of-sequence packets: 0
              	Fail Times(0/5)
              	Packet sent: 550, received: 540, Sequence(sent/rcvd/exp): 551/551/552
```

And API:

```json
      "1":{
        "vlan2010":{
          "status":"up",
          "latency":10.788033485412598,
          "jitter":9.6377668380737305,
          "packet_loss":3,
          "packet_sent":603,
          "packet_received":593,
          "session":1,
          "tx_bandwidth":22510,
          "rx_bandwidth":35201,
          "state_changed":1613409587
        }
      }
```

So it seems jitter is in ms as well. I increased the jitter to 50 ms on the Linux side and it caused the avg jitter on the Fortigate side to be reported as 25 ms. I think that's a bit weird accounting on the Fortigate side, but that is nothing we can (or should) do about that. 

`packet_loss` is definitely in percentage though (0-100).

README.md
```diff
@@ -54,6 +54,18 @@ Per-VDOM:
  * `fortigate_ha_member_bytes_total`
  * `fortigate_ha_member_ips_events_total`
 
+ Per-Link and VDOM:
```
bluecmd, line 57: This list needs to be updated as well.

probe.go
```diff
@@ -670,6 +670,121 @@ func probeLicenseStatus(c FortiHTTP) ([]prometheus.Metric, bool) {
...
+			"signals if the status is up (1) or down (0)",
+			[]string{"vdom", "link_group", "link"}, nil,
+		)
+		linkLatency = prometheus.NewDesc(
+			"fortigate_link_latency_seconds",
+			"latency of this link in seconds",
+			[]string{"vdom", "link_group", "link"}, nil,
+		)
+		linkJitter = prometheus.NewDesc(
+			"fortigate_link_jitter_seconds",
```
bluecmd, line 686: Given the discussion, WDYT about `fortigate_link_latency_jitter_seconds`?

probe.go
```diff
@@ -670,6 +670,121 @@ func probeLicenseStatus(c FortiHTTP) ([]prometheus.Metric, bool) {
...
+			"Electronic jitter on this link in seconds",
+			[]string{"vdom", "link_group", "link"}, nil,
+		)
+		linkPacketLoss = prometheus.NewDesc(
+			"fortigate_link_packet_loss_percentage",
```
bluecmd, line 691: Prometheus [suggests ](https://prometheus.io/docs/practices/naming/) `_ratio` (0.0-1.0) as the base unit for percentage. I think we should change this to be `x / 100.0` and use `_ratio` to be more conformant.

probe.go
```diff
@@ -670,6 +670,121 @@ func probeLicenseStatus(c FortiHTTP) ([]prometheus.Metric, bool) {
 	return m, true
 }
 
+func probeLinkMonitor(c FortiHTTP) ([]prometheus.Metric, bool) {
+	var (
+		linkStatus = prometheus.NewDesc(
+			"fortigate_link_status",
+			"signals if the status is up (1) or down (0)",
+			[]string{"vdom", "link_group", "link"}, nil,
```
bluecmd, line 678: I would like to expose that these metrics come from the "link monitor" subsystem somehow. Adding it to the metric name seems wrong, it would be too long and annoying. I was thinking of that maybe we can rename the `link_group` to be `monitor` - that way it would be something like:
`fortigate_link_latency_seconds{monitor="google-dns",...} 0.001`

I think that looks pretty neat, what do you think?

probe.go
secustor: yeah, that makes it clearer

probe.go
secustor: locally already fixed ;)

probe.go
secustor: I have no direct access to a fortigate so some of the terminology is a mystery for me.   If it is called like that in the UI this would be a better fit.

probe.go
secustor: Are the averages always for 500ms like your screenshots suggest or is this some setting on the monitor?

bluecmd: It is the probe interval it seems. I changed it by `set interval 750` and now it says `Interval: 750 ms`.

The avg is probably based on this:
```
ftg01 (1) # set 
[...]
probe-count                 Number of most recent probes that should be used to calculate latency and jitter (5 - 30, default = 30).
[..]
```

bluecmd:
Let me know when you are ready for the next review, I think I might have been too quick to "help" you by adding the _bps - feel free to force-push over it.

I feel it is super-close to merge this, so I get a bit jumpy :-).

secustor:
I have tried to find some best practices for unit suffixes, prometheus.io suggets to use `_byte_per_second` instead of `bps`. https://prometheus.io/docs/practices/naming/#base-units

bluecmd:
> I have tried to find some best practices for unit suffixes, prometheus.io suggets to use `_byte_per_second` instead of `bps`. https://prometheus.io/docs/practices/naming/#base-units

Good idea, as `_bps` Prometheus is always lower-case it can be both Bps (bytes per second) or bps (bits per second).
It seems the `tx_bandwidth` is **bits per second** from the UI and the CLI, unless they are showing "bps" and incorrectly assume bytes per second.

secustor:
@bluecmd I have force pushed the changes

bluecmd:
Did you want to change `fortigate_link_packet_loss_percentage` to `fortigate_link_packet_loss_ratio` ?

secustor:
Haven't planned it, but you are right that is also suggested

secustor:
When you are ready to merge, I will squash the commits

bluecmd:
Change `fortigate_link_jitter_seconds` to `fortigate_link_latency_jitter_seconds` as we discussed. I think you missed this (unless I misunderstood you?).

`fortigate_link_status_changes_total` seems to be a timestamp when the last change was, not the count of them (1.6 billion changes seems quite high, and given that `1613417367` is the current UNIX timestamp, it seems sensible that it is a timestamp. Changing it to be `fortigate_link_last_status_change` as a gauge (and the description accordingly) would probably make sense.

secustor:
Yeah I have thought we only change the description. 

What do you think about `fortigate_link_last_status_change_time_seconds` seems to fit to the base unit guide and have to be gauge based on this guide https://www.robustperception.io/are-increasing-timestamps-counters-or-gauges

bluecmd:
Ah I see. Would you be OK changing the name as well to `fortigate_link_latency_jitter_seconds` or what do you think? I want to emphasize that the two latency metrics belong together.

I had a look at Prometheus internal metrics and it has `go_memstats_last_gc_time_seconds` so `_time_seconds` seems to be fine. I think we can drop the `_last_` part as that is implied as it is a timestamp, and it is long enough as it is.  So `fortigate_link_status_change_time_seconds`. WDYT?

secustor:
I have implemented the two changes

bluecmd (approved):

bluecmd:
LGTM! You mentioned you wanted to squash it so I will let you do it.

secustor:
Done

bluecmd:
Thanks a million!

Issue #38, opened 2021-02-22
Provide Release artifacts and pre build Docker images


secustor:
To enable a easier start up for user we should pre-build Docker images and go binaries.    

This could be in the first step a simple image stream which follows the master branch and later on a full blown automatic releases e.g. with [semver-bot](https://github.com/semantic-release/semantic-release)

bluecmd:
I set up quay.io for now as I have used it in the past.

We can switch to use something else if there is a community preference.

https://quay.io/repository/bluecmd/fortigate_exporter?tab=info

secustor:
semantic-release would create following release: 

# 1.0.0 (2021-03-14)

### Bug Fixes

    * CI: semver configuration typo (d3c85ff (https://github.com/bluecmd/fortigate_exporter/commit/d3c85ff5d75d93774d5e830a80c4ccade9b44dba))
    * deps: update module github.com/prometheus/client_golang to v1.9.0 (d91e26d (https://github.com/bluecmd/fortigate_exporter/commit/d91e26d9e3f2725a863689441119d9a767d7320e))
    * deps: update module gopkg.in/yaml.v2 to v2.4.0 (44c6b4e (https://github.com/bluecmd/fortigate_exporter/commit/44c6b4e95c775bf2872c2640a8767c0de0ad4fa6))
    * probeVirtualWANPerf: change Array type from string to float64 (0ec28d1 (https://github.com/bluecmd/fortigate_exporter/commit/0ec28d1587c6215c7bbce747a888450417fff8a4))
    * add api path to ha-statistics.jsonnet (e17a64d (https://github.com/bluecmd/fortigate_exporter/commit/e17a64d555bbbf4abaf0a9cf81f4a7d6be191466))

### Features

    * CI: add semantic-release for auto creation of releases, for now in dryRun mode (4125c35 (https://github.com/bluecmd/fortigate_exporter/commit/4125c350fa0ceb234c2f7091fc9e2533f249d479))
    * CI: replace composite actions with inline steps (2f62f83 (https://github.com/bluecmd/fortigate_exporter/commit/2f62f832f133bf43c3c54b3c794bd5ebde002380))
    * CI: run go workflow on all branches except master (14f6979 (https://github.com/bluecmd/fortigate_exporter/commit/14f6979f7a383e9f1190197bca67a6b21c388ef8))
[...]

PR #48, opened 2021-03-02
doc: instructions for usage of CI

bluecmd:
Add instructions how to get started using the automatic builds.

Ping #38

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,7 @@
 
 ![Go](https://github.com/bluecmd/fortigate_exporter/workflows/Go/badge.svg)
 ![Docker](https://github.com/bluecmd/fortigate_exporter/workflows/Docker/badge.svg)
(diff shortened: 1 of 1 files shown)


secustor (commented):
Is it intended that the build is run for all branches?    

I have never used quay.io for building.  What happens if the unit tests are failing?      
They are probably build anyway right?

bluecmd:
> Is it intended that the build is run for all branches?
> 
> I have never used quay.io for building. What happens if the unit tests are failing?
> They are probably build anyway right?

We can disable all branches if you want - I think it should only build for the branches we have in the repository, which only we and Renovate can create.

My thinking was that if we enable building on all branches, it would be easy for people to try out some pending PR we have as there are images built for those. :latest is definitely going to be very confusing to use, so I guess people would have to use :master explicitly.

secustor (approved):
Some of the build jobs have failed. Is there a possibility to restart them?

bluecmd:
Seems we ran out of build quota, so I changed it to only build the master branch for now.

PR #51, opened 2021-03-02
feat(Github Workflow): check if GO files are formatted

### LINKED ISSUE #50: Add check to ensure correct code formating
To ensure that correct formatted code is merged we should add a github action which runs 

```bash 
if [ "$(gofmt -s -l . | wc -l)" -gt 0 ]; then
    exit 1
fi
```
Which will fail if a file is not properly formatted

secustor:
Closes #50

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -22,6 +22,14 @@ jobs:
     - name: Check out code into the Go module directory
       uses: actions/checkout@v2
 
+    - name: Ensure that all files are properly formatted
+      run: |
+        FILES=$(gofmt -s -l .)
+        if [ -n "${FILES}" ]; then
+            printf "Following files are not formatted: \n%s" "$FILES"
+            exit 1
+        fi
+
     - name: Get dependencies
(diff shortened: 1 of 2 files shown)


.github/workflows/go.yml
```diff
@@ -22,6 +22,14 @@ jobs:
     - name: Check out code into the Go module directory
       uses: actions/checkout@v2
 
+    - name: Ensure that all files are properly formatted
+      run: |
+        FILES=$(gofmt -s -l .)
+        if [ "$(gofmt -s -l . | wc -l)" -gt 0 ]; then
```
bluecmd, line 28: This will run gofmt twice, right? Can we do `[ ! -z "${FILES}" ];` maybe?

bluecmd (changes requested):

.github/workflows/go.yml
secustor: yeah that is definitely better.

bluecmd (approved):

PR #52, opened 2021-03-03
add metrics for load balance servers

### LINKED ISSUE #47: Include metrics for load balance servers
Feature request to expose load balance metrics (virtual servers and real servers).

api endpoint: `/firewall/load-balance`

Expected metrics (proposal):
>  Per-VirtualServer and VDOM:
> * `fortigate_lb_virtualserver_info`

> Per-RealServer for each VirtualServer and VDOM:
> * `fortigate_lb_realserver_info`
> * `fortigate_lb_realserver_mode`
> * `fortigate_lb_realserver_status`
> * `fortigate_lb_realserver_active_sessions`
> * `fortigate_lb_realserver_bytes_processed`

I am working on adding these metrics and opening a PR for it.

Just one question about how to handle the mandatory 'count' param for this endpoint.
> count *required integer(query) | Maximum number of entries to return.

Is it acceptable to just use a fixed value, e.g. 999? (is this enough? people with more than 999 virtual servers configured?, make it 9999?)
Or should it be a complex batching solution, using the start and count params, and quering in batches of e.g. 250.
```
/firewall/load-balance?start=0&count=250
/firewall/load-balance?start=250&count=250
etc
```

Any ideas on your part are welcome.

sboschman:
add load balance server (virtual servers and real servers) metrics

fixes #47


probe.go
```diff
@@ -1011,6 +1011,127 @@ func probeCertificates(c FortiHTTP) ([]prometheus.Metric, bool) {
...
+				m = append(m, prometheus.MustNewConstMetric(realServerMode, prometheus.GaugeValue, realServerModeDisabled, r.VDOM, virtualServer.Name, fmt.Sprintf("%d", realServer.ID), "disabled"))
+				m = append(m, prometheus.MustNewConstMetric(realServerStatus, prometheus.GaugeValue, realServerStatusUp, r.VDOM, virtualServer.Name, fmt.Sprintf("%d", realServer.ID), "up"))
+				m = append(m, prometheus.MustNewConstMetric(realServerStatus, prometheus.GaugeValue, realServerStatusDown, r.VDOM, virtualServer.Name, fmt.Sprintf("%d", realServer.ID), "down"))
+				m = append(m, prometheus.MustNewConstMetric(realServerStatus, prometheus.GaugeValue, realServerStatusUnknown, r.VDOM, virtualServer.Name, fmt.Sprintf("%d", realServer.ID), "unknown"))
+				m = append(m, prometheus.MustNewConstMetric(realServerSessions, prometheus.GaugeValue, float64(realServer.ActiveSessions), r.VDOM, virtualServer.Name, fmt.Sprintf("%d", realServer.ID)))
+				m = append(m, prometheus.MustNewConstMetric(realServerBytesProcessed, prometheus.CounterValue, float64(realServer.BytesProcessed), r.VDOM, virtualServer.Name, fmt.Sprintf("%d", realServer.ID)))
```
secustor, line 1128: Please use `strconv.Itoa()` instead of `printf` and remove the float conversion as with the above changes it will not be necessary anymore
```suggestion
				m = append(m, prometheus.MustNewConstMetric(realServerBytesProcessed, prometheus.CounterValue, realServer.BytesProcessed, r.VDOM, virtualServer.Name, strconv.Itoa(realServer.ID)))
```

secustor (changes requested):
I have taken a quick peak on this PR even if it is not yet marked as `ready for review`.    
If there still bigger changes simply ignore it.

probe.go
```diff
@@ -1011,6 +1011,127 @@ func probeCertificates(c FortiHTTP) ([]prometheus.Metric, bool) {
...
+
+	type RealServer struct {
+		IP     string `json:"real_server_ip"`
+		Port   int    `json:"real_server_port"`
+		ID     int    `json:"real_server_id"`
+		Mode   string `json:"mode"`
+		Status string `json:"status"`
+		// MonitorEvents  int    `json:"monitor_events"`
+		ActiveSessions int `json:"active_sessions"`
+		// How to 'parse' this field (e.g. '<1' and expose it as a usable metric?)
```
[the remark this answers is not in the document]
sboschman: No idea about possible values and unit to be honest, I can't really find a definite format for this field. You might be right that they only use '<1' and otherwise a ms value. Maybe we should just go with that assumption, if/when someone spots a different 'value' in the wild we can add it.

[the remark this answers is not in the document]
secustor: @bluecmd can you look into this with your fortigate instance? 

If we can not find any reference this will probably be the best.     
A parsing failure should not lead to other missing metrics and should be definitely logged as an error.

probe.go
```diff
@@ -1011,6 +1012,144 @@ func probeCertificates(c FortiHTTP) ([]prometheus.Metric, bool) {
...
+
+				realServerRTTValue := -1.0
+				if "<1" == realServer.RTT {
+					realServerRTTValue = 1.0
+				} else {
+					var err error
+					realServerRTTValue, err = strconv.ParseFloat(realServer.RTT, 64)
+					if err != nil {
+						log.Printf("Failed to parse RTT value: %v", err)
+						realServerRTTValue = -1.0
```
secustor, line 1133: Please only export the metric if there is no error.     
@sboschman or what have been your thoughts regarding this?

secustor (changes requested):

probe.go
sboschman: I was thinking it might be handy to tell something is wrong by the metric value, prompting you to have a look in the logs. Instead of having a sporadic missing metric. Assume if the RTT is exactly 1 sec, it isn't reported as 1000 but as 1s.

On the other hand, a (sporadic) missing metric should perhaps be enough to prompt an inspection of the logs.

secustor: Ah, I have overlooked that one, the time values should definitely be `seconds` . https://prometheus.io/docs/practices/naming/#base-units     

@bluecmd  WDYT regarding the dropping of the metric.  
ATM we have three options: 
- set the value `-1`, like it is implemented right now
- dropping the metric in the failure case
- dropping the metric in the failure case and return a non success

bluecmd: I'd say let's use NaN instead for the error case. It works better using things like `sum` and `avg` etc and will not be lost as noise when people are looking at the metrics in aggregate (as they likely will).

Dropping the metric has the issue that Prometheus does stale metric caching, so it will effectively retain the older value for queries upon it. A NaN will look very odd and cause people to dig deeper, I think.

bluecmd: And yes, please expose it as _seconds as per the Prometheus base units as @secustor said.

sboschman: NaN :+1:

secustor (approved):

secustor:
LGTM, thx for your contribution!

PR #53, opened 2021-03-03
use gock for HTTP mocking

sboschman:
Replace the fakeClient impl with gock. This should open up an easier way to introduce pagination to certain FortiAPI calls, as only one client impl has to be changed/extended to support pagination.

Related to #52 , as the api call used to get load balance servers has a mandatory count param for pagination


bluecmd:
Closing due to inactivity

Issue #56, opened 2021-03-11
 Error: json: cannot unmarshal number 64687125982 into Go struct field pStats.Results.Bytes of type int


lukaszsedek:
When I run fortigate_exporter I see this error:

`2021/03/11 08:50:12 Error: json: cannot unmarshal number 64687125982 into Go struct field pStats.Results.Bytes of type int
2021/03/11 08:50:13 Probe of "https://192.168.30.1" failed, took 0.766 seconds`

bluecmd:
That value is 36 bits it seems, so likely the int type defaults to 32-bit for this particular build.

Changing to int64 might be the correct thing here, worst case float64. Should be easy to repro in test cases.

Thanks for the report!

bluecmd:
@lukaszsedek can you try with the latest version and see if the issue is gone for you?

secustor:
Closing this issue as we have merged code which should fix this and haven't received an answer regarding the current state of the issue.   

If the issue is still their, feel free to reopen this issue.

PR #57, opened 2021-03-11
feat: use float64 or int64 to prevent errors from too big numbers 

secustor:
This PR replaces ints used for parsing with int64 and float64 to prevent that too big numbers are tried to be parsed.     
Further it reduces the parsing steps through signalling the parser to directly parse to float64 instead of parsing it in an extra step. 

I couldn't reproduce the error but this should be IMHO done anyway. 
Context: #56

### DIFF
--- a/probe.go
+++ b/probe.go
@@ -48,7 +48,7 @@ func probeSystemStatus(c FortiHTTP) ([]prometheus.Metric, bool) {
 		Status  string
 		Serial  string
 		Version string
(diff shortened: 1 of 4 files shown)


bluecmd:
What do you think about adding a test on at least one metric to test values >32 bit?

secustor:
I have added two unit test for this.

bluecmd (approved):

PR #63, opened 2021-03-14
Create EXAMPLES.md

### LINKED ISSUE #39: Add example PromQl queries to README.md
Add some PromQL queries/rules for the Fortigate exporter metrics to show some of the use cases which are covered.    

The metric names alone do not allow to visualize most use cases if you are not deeply invested in the Prometheus ecosystem.  
This would allow new users to get better earlier results and therefore more users in the long run. 

Later on the scope can be widened to a [Mixin](https://github.com/kubernetes-monitoring/kubernetes-mixin)  solution. e.g. best practice alarming rules and dashboards like they are provided for Kubernetes and the Prometheus stack itself

bluecmd:
Add examples how to use the Fortigate Exporter

Fixes #39 but of course more examples will be added over time.

### DIFF
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -7,7 +7,7 @@ To help the users of this exporter, here are some examples to illustrate
 how one could use the power of PromQL to query interesting facts about one's
 environment.
 
-For an introduction to PromQL, see the official Prometheus
+For an introduction to the PromQL basics, see the official Prometheus
 [documentation](https://prometheus.io/docs/prometheus/latest/querying/examples/).
 
 ## Most Active Firewall Policies
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -1,0 +1,60 @@
+# PromQL usage examples for Fortigate Exporter
+
+PromQL is an advanced language and proper usage of it requires one to both understand
+the language itself and how the metrics one has are constructed.
+
+To help all the users of this exporter, here are some examples to illustrate
+how one could use the power of PromQL to query interesting facts about one's
+environment.
+
+## Most Active Firewall Policies
+
+Using the `topk` function it is easy to get the most active firewall policies
+based on whatever ranking function you want.
+
+Example: `topk(3, rate(fortigate_policy_bytes_total[15m])) * 8`
+This will return the top 3 most active policies based upon the bytes transfered on
+average the last 15m. The output will be bits/s.
+
+| Element | Value| 
+|---------|------|
+`{id="9",instance="fgt-a",job="fortigate",name="SSH",protocol="ipv6",uuid="5cd4b62e-4904-51eb-b4a9-f52e75461e52",vdom="bluecmd"}` |	102805.48826815643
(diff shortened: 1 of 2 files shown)


EXAMPLES.md
```diff
@@ -0,0 +1,60 @@
+# PromQL usage examples for Fortigate Exporter
+
+PromQL is an advanced language and proper usage of it requires one to both understand
+the language itself and how the metrics one has are constructed.
+
+To help all the users of this exporter, here are some examples to illustrate
+how one could use the power of PromQL to query interesting facts about one's
+environment.
```
secustor, line 8: Maybe we should here a link to the PromQL documentation?   https://prometheus.io/docs/prometheus/latest/querying/examples/

secustor (commented):
What do you think about moving this in a docs folder?

bluecmd:
> What do you think about moving this in a docs folder?

Hm, will we have many more?
I am thinking README.md, CONTRIBUTING.md, and EXAMPLES.md.
The first two will need to be in the root because of well, that's where people expect them to be.

I'd say we do not create a docs/ for now but I am 100% with you to move it if we end up having a website or something.

bluecmd:
Ready for review. Had a friend (thank @ventris) look over it quite quickly as well to see that it reads OK.

EXAMPLES.md
```diff
@@ -0,0 +1,67 @@
+# PromQL usage examples for Fortigate Exporter
+
+PromQL is an advanced language and proper usage of it requires one to both understand
+the language itself and how the metrics one has are constructed.
```
secustor, line 4: English is my second language so could be me, but that sounds kinda weird. 

WDYT about  `PromQL is an advanced language and proper usage of it requires first to understand the language itself and how the metrics one are constructed.`

EXAMPLES.md
```diff
@@ -0,0 +1,67 @@
...
+
+To help the users of this exporter, here are some examples to illustrate
+how one could use the power of PromQL to query interesting facts about one's
+environment.
+
+For an introduction to PromQL, see the official Prometheus
```
secustor, line 10: Maybe?
```suggestion
For an introduction to the PromQL basics, see the official Prometheus
```

EXAMPLES.md
bluecmd: Yeah, it is a hard sentence to grok. What do you think of this new one?

secustor: What are trying to bring across with organized? I'm not sure what you mean here

bluecmd: I want to state that even if you are the world's best champion on PromQL, you still need to learn how each exporter works; as in what metrics are available to you, and how their labels associate between the metrics.

E.g. that we have a fortigate_version_info that can be used to get the build info (as opposed to adding the build label to all metrics, as another exporter might have had done).

secustor: I see. :)

secustor (approved):

Issue #67, opened 2021-03-14
GO binaries haven't been added to the release


secustor:
The build binaries of the FortigateExporter have not been added to the Release.

bluecmd:
```
[10:08:17 PM] [semantic-release] › ✖  ENOGHTOKEN No GitHub token specified.
A GitHub personal token (https://github.com/semantic-release/github/blob/master/README.md#github-authentication) must be created and set in the GH_TOKEN or GITHUB_TOKEN environment variable on your CI environment.

Please make sure to create a GitHub personal token (https://help.github.com/articles/creating-a-personal-access-token-for-the-command-line) and to set it in the GH_TOKEN or GITHUB_TOKEN environment variable on your CI environment. The token must allow to push to the repository bluecmd/fortigate_exporter.

[10:08:17 PM] [semantic-release] › ✖  ENOGHTOKEN No GitHub token specified.
A GitHub personal token (https://github.com/semantic-release/github/blob/master/README.md#github-authentication) must be created and set in the GH_TOKEN or GITHUB_TOKEN environment variable on your CI environment.

Please make sure to create a GitHub personal token (https://help.github.com/articles/creating-a-personal-access-token-for-the-command-line) and to set it in the GH_TOKEN or GITHUB_TOKEN environment variable on your CI environment. The token must allow to push to the repository bluecmd/fortigate_exporter.

AggregateError: 
    SemanticReleaseError: No GitHub token specified.
        at module.exports (/home/runner/.npm/_npx/6828/lib/node_modules/semantic-release/node_modules/@semantic-release/github/lib/get-error.js:6:10)
        at module.exports (/home/runner/.npm/_npx/6828/lib/node_modules/semantic-release/node_modules/@semantic-release/github/lib/verify.js:96:17)
        at verifyConditions (/home/runner/.npm/_npx/6828/lib/node_modules/semantic-release/node_modules/@semantic-release/github/index.js:27:9)
        at validator (/home/runner/.npm/_npx/6828/lib/node_modules/semantic-release/lib/plugins/normalize.js:34:30)
        at /home/runner/.npm/_npx/6828/lib/node_modules/semantic-release/lib/plugins/pipeline.js:37:40
        at next (/home/runner/.npm/_npx/6828/lib/node_modules/semantic-release/node_modules/p-reduce/index.js:17:9)
    at /home/runner/.npm/_npx/6828/lib/node_modules/semantic-release/lib/plugins/pipeline.js:54:11
[...]

PR #69, opened 2021-03-14
fix(CI): Add GitHub environment

bluecmd:
This might help #67 - but the documentation is not clear.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,6 +9,7 @@ jobs:
   release:
     name: 'Release to GitHub'
     runs-on: ubuntu-latest
+    environment: GH_TOKEN
     steps:
       - uses: actions/checkout@v2
 # Go setup


bluecmd:
I added a GH_TOKEN secret but it still does not seem to work. This is how the secret page looks like:

image

It seems the Environment stuff is pretty new? And to activate it for an action it seems that we need to do something like this.

This is what the environment says:

image

secustor:
Strange. The environment variable get set here  https://github.com/bluecmd/fortigate_exporter/blob/9c18f19e953a37e3653c47911ea46b7cfe7806ec/.github/workflows/release.yml#L57

This seems correct based on this documentation: https://docs.github.com/en/actions/reference/encrypted-secrets#using-encrypted-secrets-in-a-workflow and this blog https://www.asyncapi.com/blog/automated-releases-part-two

secustor:
Plugin docs: https://github.com/semantic-release/github#github-authentication

bluecmd:
image

This seems to be a better configuration, now things seem to work as far as the secret goes. Still no binaries though, but this PR is not needed.

PR #74, opened 2021-03-19
feat: Implement versioning and add Makefile

### LINKED ISSUE #61: Add build-time information as metric and in logs
We should expose what git commit was used to build the currently running exporter, as well as any tag associated with it. That way it is easier to reason about what version was running both in the logs and in metrics.

secustor:
Fixes #61

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,20 +1,31 @@
-VERSION := $(shell git tag | grep ^v | sort -V | tail -n 1)
+VERSION := $(shell git describe --tags)
 GIT_HASH := $(shell git rev-parse --short HEAD )
-LDFLAGS = -ldflags "-X main.Version=${VERSION} -X main.GitHash=${GIT_HASH}"
 
+GO_VERSION        ?= $(shell go version)
+GO_VERSION_NUMBER ?= $(word 3, $(GO_VERSION))
+LDFLAGS = -ldflags "-X main.Version=${VERSION} -X main.GitHash=${GIT_HASH} -X main.GoVersion=${GO_VERSION_NUMBER}"
+
+.PHONY: build
 build:
 	go build ${LDFLAGS} -v -o target/fortigate-exporter .
 
+.PHONY: build-release
 build-release:
 	GOOS=linux   GOARCH=amd64 go build ${LDFLAGS} -o=target/fortigate-exporter.linux.amd64 .        && \
   	GOOS=windows GOARCH=amd64 go build ${LDFLAGS} -o=target/fortigate-exporter.windows.amd64.exe .  && \
   	GOOS=darwin  GOARCH=amd64 go build ${LDFLAGS} -o=target/fortigate-exporter.darwin.amd64 .
 
+.PHONY: test
 test:
 	go test -v .
 
+.PHONY: get-dependencies
 get-dependencies:
 	go get -v -t -d ./...
 
+.PHONY: vet
 vet:
 	go vet ./...
+
+test-output:
+	$(shell echo $$GO_VERSION_NUMBER)
--- a/Makefile
+++ b/Makefile
@@ -11,9 +11,9 @@ build:
 
 .PHONY: build-release
 build-release:
-	GOOS=linux   GOARCH=amd64 go build ${LDFLAGS} -o=target/fortigate-exporter.linux.amd64 .        && \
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -0,0 +1,20 @@
+VERSION := $(shell git tag | grep ^v | sort -V | tail -n 1)
+GIT_HASH := $(shell git rev-parse --short HEAD )
+LDFLAGS = -ldflags "-X main.Version=${VERSION} -X main.GitHash=${GIT_HASH}"
+
```
bluecmd, line 4: Please add .PHONY for the pseudo-targets (https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html)

Makefile
```diff
@@ -0,0 +1,20 @@
+VERSION := $(shell git tag | grep ^v | sort -V | tail -n 1)
```
bluecmd, line 1: Maybe `git describe --tags` is better? That will show the current tag on the current branch - which means you can do things like:

```
$ git checkout v1.0.1
$ make
```

If I read correctly, the above will always return the latest version in the git repository, unrelated to what commit you are building from.

fortigate_exporter.go
```diff
@@ -54,6 +55,18 @@ type FortiHTTP interface {
 	Get(path string, query string, obj interface{}) error
 }
 
+func setUpMetricsEndpoint(version string, hash string) {
+	fortigateExporterInfo := promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "fortigate_exporter_info",
```
bluecmd, line 60: fortigate_exporter_build_info maybe?
It seems like this might be a bit _too_ generic right now.

fortigate_exporter.go
```diff
@@ -118,6 +131,10 @@ func probeHandler(w http.ResponseWriter, r *http.Request) {
 func main() {
 	flag.Parse()
 
+	version, hash := getVersion()
+	log.Printf("FortigateExporter:\tversion: %s\t\tcommit: %s", version, hash)
```
bluecmd, line 135: A bit unrelated, but I would like us to switch to using uber's Zap logging in the future to allow for optional structured logs - the normal human logs are much nicer as well. This is fine for now though.

version.go
```diff
@@ -33,3 +34,20 @@ func ParseVersion(ver string) (int, int, bool) {
 	}
 	return major, minor, true
 }
+
+var (
+	Version = "(devel)"
+	GitHash = "(no hash)"
+)
+
+func getVersion() (string, string) {
```
bluecmd, line 43: version.go was meant to parse Fortigate version numbers, so it probably makes more sense to keep this function in the fortigate_exporter.go file.

bluecmd (changes requested):

Makefile
secustor: k, will add this. This is my first Makefile so the learning curve is pretty steep.

fortigate_exporter.go
secustor: Hmm I have looked into what other projects are using here some examples: 
```
grafana_build_info{branch="HEAD",edition="oss",goversion="go1.15.6",revision="c2203b9859",version="7.4.0"} 1
prometheus_build_info{branch="HEAD", goversion="go1.15.3", revision="0a7fdd3b76960808c3a91d92267c3d815c1bc354",  version="2.22.0"
```
I will change this to mimic other projects

secustor: Regarding the application version. Should we strip the v from the tags, as Grafana and Prometheus do?     
So, basically `v1.1.1` vs `1.1.1` .

bluecmd: Yes, let's strip the v if that is what the others do.

secustor:
I had to move the building of the artifacts into semantic-release, as we have here a race condition between the tag creation and the compiling.

fortigate_exporter.go
```diff
@@ -44,16 +47,57 @@ var (
...
+}
+
 type FortiHTTP interface {
 	Get(path string, query string, obj interface{}) error
 }
 
+func setUpMetricsEndpoint(buildInfo BuildInfo) {
+	fortigateExporterInfo := promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "fortigate_exporter_build_info",
+		Help: "This info metric contains generic information for about the exporter",
```
bluecmd, line 72: The help seems to be outdated now.

bluecmd (approved):

PR #81, opened 2021-04-07
fea(ci): Disable CGO for building

### LINKED ISSUE #80: Release binaries should not be built using CGO
Right now they seem to be linked using CGO and result in dynamically linked binaries:

```
$ file fortigate-exporter.linux.amd64
fortigate-exporter.linux.amd64: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, Go BuildID=aC_NDJcH75wKecO7Svr3/siFCDbSCqhv7y72S7Mz7/wOdLiG5GObmgWQWsk5Vj/M1trWEvxBKJed0WeW5tI, not stripped
$ ldd fortigate-exporter.linux.amd64
        linux-vdso.so.1 (0x00007fffd2a3e000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f4f1919e000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f4f18fd0000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f4f191db000)
```

Building with `CGO_ENABLED=0` seems to work as far as I can see, so I will submit a PR for that.

bluecmd:
This should result in static binaries, and we should not need CGO
anyhow.

Fixes #80

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,14 +14,14 @@ build-release: build-release-amd64 build-release-arm64
 
 .PHONY: build-release-amd64
 build-release-amd64:
-	GOOS=linux   GOARCH=amd64 go build ${LDFLAGS} -o=fortigate-exporter.linux.amd64 .        && \
-  	GOOS=windows GOARCH=amd64 go build ${LDFLAGS} -o=fortigate-exporter.windows.amd64.exe .  && \
-  	GOOS=darwin  GOARCH=amd64 go build ${LDFLAGS} -o=fortigate-exporter.darwin.amd64 .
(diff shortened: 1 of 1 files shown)


secustor (changes requested):
LGTM except change `fea` with `feat` in your commit messge.

bluecmd:
Done!

secustor (approved):

Issue #83, opened 2021-04-08
Installation


stevelogik:
Hello!

I'd like some help and instruction on installation. Via the installation instructions, I found below, is this for installing this on a linux machine? What would be the best way to download the binary on Linux for installation?

Download the newest binary for your system https://github.com/bluecmd/fortigate_exporter/releases/latest. Be aware that most modern computers will use AMD64, but ARM64 is used more and more too, e.g. a RaspberryPi

Set up the monitors and permission on your Fortigate device https://github.com/bluecmd/fortigate_exporter/tree/master#fortigate-configuration

create a API Token on your Fortigate device. Note the API Token for the next step.

create a `fortigate-key.yaml` file next to the binary you have downloaded

Add the URL of your Fortigate as key and the API token as value

Start the Exporter and test it using curl like described in https://github.com/bluecmd/fortigate_exporter/tree/master#usage

bluecmd:
Hello!

> I found below, is this for installing this on a linux machine?

Yes

> What would be the best way to download the binary on Linux for installation?

On Linux? Probably something like this in your terminal:
```
wget https://github.com/bluecmd/fortigate_exporter/releases/download/v1.4.0/fortigate-exporter.linux.amd64
chmod +x fortigate-exporter.linux.amd64
```

Create the file named for example `fortigate-key.yaml`, then again in your terminal:
```
./fortigate-exporter.linux.amd64 -auth-file fortigate-key.yaml
```

stevelogik:
docker build and docker run commands I can run on a fresh linux install? Can I set the fortigate-exporter to run on startup?

Also, I see the command "./fortigate_exporter -auth-file ~/fortigate-key.yaml" ; When running on windows and running this command, I'm assuming this command runs the exporter? Can you create this to run as a service? Or can you only run this in a PowerShell prompt where you need to leave the terminal open

bluecmd:
Hi.

I am sorry, but for Windows I do not know how to easy run this as a service. I am sure there are materials out there on the Internet, as this would be something that would be applicable to a lot of other software.

For Linux, you can easily create a systemd configuration file starting the fortigate exporter on startup.
I will add an systemd example to the README.

secustor:
Closing because of inactivity

PR #85, opened 2021-04-12
doc(systemd): Add example service file

bluecmd:
This is the service file I use for my own servers as a reference example.

Related to #83


secustor:
I think we should put this inside a folder, e.g. `production` or `deployments`. Ss we add additional deployment methods it will getting more and more confusing. 
Further I think there should be a reference to this file in the README

PR #86, opened 2021-04-16
Fix for ipsec with multiple P2 using the same name

per-lind:
Currently the exporter will fail if it runs into a IPSec tunnel with multiple phase2 proposals having the same name. 
This fix adds the p2serial found in the api as another label.


bluecmd (approved):
LGTM, thanks!

PR #89, opened 2021-04-21
Wifi examples

kristenungdom:


secustor (changes requested):
Please, add in the first line from which endpoint the response comes.

PR #95, opened 2021-06-15
refactor: split up probes and main package

secustor:
This PR refactors the program structure with the target to make easier to read and find specific parts of the code. 

Sub functions are capsuled inside of other packages


Makefile
```diff
@@ -25,7 +25,7 @@ build-release-arm64:
 
 .PHONY: test
 test:
-	go test -v .
+	go test -v -race ./...
```
bluecmd, line 28: 👍

Makefile
```diff
@@ -37,3 +37,7 @@ vet:
 
 test-output:
 	$(shell echo $$GO_VERSION_NUMBER)
+
+.PHONY: fmt-fix
+fmt-fix:
+	gofmt -s -w -l .
```
bluecmd, line 43: If we want to vendor (I had to do this for some stuff recently even in 1.16) this will hit us, but that might be a thing to fix then.

fortigate_exporter.go
```diff
@@ -18,54 +18,30 @@
 package main
 
 import (
-	"context"
-	"crypto/tls"
-	"crypto/x509"
-	"flag"
-	"fmt"
+	"github.com/bluecmd/fortigate_exporter/internal/config"
```
bluecmd, line 21: Is there a reason to combine the imports? I would probably say it is more established practice to split core imports and 3rd party ones.

fortigate_exporter.go
```diff
@@ -98,116 +74,29 @@ func getBuildInfo() BuildInfo {
...
 	buildInfo := getBuildInfo()
 	log.Printf("FortigateExporter %s ( %s )", buildInfo.version, buildInfo.gitHash)
 	setUpMetricsEndpoint(buildInfo)
 
-	roots, err := x509.SystemCertPool()
-	if err != nil {
-		log.Fatalf("Unable to fetch system CA store: %v", err)
+	initErr := config.Init()
+	if initErr != nil {
+		log.Fatalf("%q", initErr)
```
bluecmd, line 84:
```suggestion
	if err := config.Init(); err != nil {
		log.Fatalf("Initialization error: %v", err)
```

fortigate_exporter.go
```diff
@@ -98,116 +74,29 @@ func getBuildInfo() BuildInfo {
...
-	tc := &tls.Config{RootCAs: roots}
-	if *insecure {
-		tc.InsecureSkipVerify = true
-	}
-	http.DefaultTransport.(*http.Transport).TLSHandshakeTimeout = time.Duration(*tlsTimeout) * time.Second
-	http.DefaultTransport.(*http.Transport).TLSClientConfig = tc
 
-	af, err := ioutil.ReadFile(*authMapFile)
+	savedConfig := config.GetConfig()
+	err := fortiHttp.Configure(savedConfig)
```
bluecmd, line 88: Same here, `if err := x; er != nil { ` and not `%q` (probably `%v`)

pkg/probes/firewall_load_balance.go
```diff
@@ -0,0 +1,156 @@
+package probes
```
bluecmd, line 1: Go seems to, and I agree with it, prefer having singular for package names.
If you feel strongly about it I can let it slide, but I would prefer singular if possible.

internal/config/main.go
```diff
@@ -0,0 +1,105 @@
+package config
```
bluecmd, line 1: What do you think of `pkg/internal/config` as the path instead?

internal/config/main.go
```diff
@@ -0,0 +1,105 @@
...
+	"io/ioutil"
+	"log"
+	"strings"
+)
+
+type FortiExporterParameter struct {
+	AuthFile      *string
+	Listen        *string
+	ScrapeTimeout *int
+	TlsTimeout    *int
```
bluecmd, line 15: `TLSTimeout` etc.

bluecmd (commented):
Very good, I like this!

Of course some naming arguments as you probably expected :-).

fortigate_exporter.go
secustor: adapted the pipeline and Makefile to handle this

internal/config/main.go
secustor: Why put this in pkg?  I think there will be no value for someone external to pull this into his code

bluecmd: I was thinking so we concentrate where the go-code lives. Main directory (`*.go`) and one sub-directory, but it is fine - I looked at what Prometheus was doing and they have it spread out so can't be that bad.

fortigate_exporter.go
```diff
@@ -18,54 +18,32 @@
...
-	"net/url"
 	"runtime"
 	"runtime/debug"
 	"strings"
-	"time"
 
+	"github.com/bluecmd/fortigate_exporter/pkg/probe"
+
+	"github.com/bluecmd/fortigate_exporter/internal/config"
+	fortiHttp "github.com/bluecmd/fortigate_exporter/pkg/http"
```
bluecmd, line 30: fortiHTTP

bluecmd (approved):

internal/config/main.go
secustor: I have followed mostly the standard [Go structure](https://github.com/golang-standards/project-layout)

But I have skipped the CMD part as I think we should not change that much in a single step, to prevent unexpected problems

Issue #96, opened 2021-06-16
Error 500 when scraping 1 out of 4 Fortigate firewalls


peter-funktionIT:
Hi, 
First of all, thanks for a great exporter.

We are using this to scrape four firewalls in total, at different locations.
Three of them work fine, but one gives an error in Prometheus /targets, 
`server returned HTTP status 500 Internal Server Error`

Let's call the working ones fw1, fw2 and fw3, and the non-working one fw4

From the Prometheus server I can scrape this firewall manually just fine:
`curl -k 'https://<ip of fw4>:8443/api/v2/cmdb/firewall/address?vdom=root&access_token=<redacted>'`
Output:
```
{
  "http_method":"GET",
  "revision":"3ebf11d4279c8c3450fa3c22132ffa07",
  "results":[
    {
<cut to the end>
    }
  ],
  "vdom":"root",
  "path":"firewall",
  "name":"address",
  "status":"success",
  "http_status":200,
  "serial":"<redacted>",
  "version":"v6.2.4",
  "build":1112
}
```

Fortigate exporter logs reports it can scrape just fine:
```
Jun 16 11:50:35 prometheus01 fortigate_exporter[3217205]: 2021/06/16 11:50:35 Probe of "https://<ip of fw4>:8443" succeeded, took 0.305 seconds
Jun 16 11:50:47 prometheus01 fortigate_exporter[3217205]: 2021/06/16 11:50:47 Probe of "https://<ip of fw1>:8443" succeeded, took 0.345 seconds
Jun 16 11:50:54 prometheus01 fortigate_exporter[3217205]: 2021/06/16 11:50:54 Probe of "https://<ip of fw3>:8443" succeeded, took 0.381 seconds
Jun 16 11:50:59 prometheus01 fortigate_exporter[3217205]: 2021/06/16 11:50:59 Probe of "https://<ip of fw2>:8443" succeeded, took 0.447 seconds
Jun 16 11:51:05 prometheus01 fortigate_exporter[3217205]: 2021/06/16 11:51:05 Probe of "https://<ip of fw4>:8443" succeeded, took 0.311 seconds

```

There are no logs in Prometheus indicating an error.

Scrape config in prometheus should be identical for all four firewalls:
```
scrape_configs:
  - job_name: fortigate_exporter
    metrics_path: /probe
    relabel_configs:
    - source_labels:
      - __address__
      target_label: __param_target
    - regex: (?:.+)(?::\/\/)([^:]*).*
[...]

PR #101, opened 2021-07-09
test(ipsec): Test behaviour on common phase2 IDs

bluecmd:
This tests that issue #96 was fixed by #86.


secustor (approved):
LGTM but maybe use `test` commitType as suggested by the conventional commits guideline

PR #110, opened 2021-07-10
feat(flag): Valid default for auth-file

bluecmd:
Use the basename of the configuration file we use by default in the Docker image.

This avoids this non-helpful error message when running without arguments:
```
[debian]$ ./fortigate-exporter.linux.amd64
2021/07/10 21:45:43 FortigateExporter 1.5.1-1-g9994480 ( 9994480 )
2021/07/10 21:45:43 Failed to read API authentication map file: open : no such file or directory
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,7 +137,7 @@ To probe a FortiGate, do something like `curl 'localhost:9710/probe?target=https
 ### Available CLI parameters
 | flag  | default value  |  description  |
 |---|---|---|
-| -auth-file      | /config/fortigate-key.yaml  | path to the location of the key file |
+| -auth-file      | fortigate-key.yaml  | path to the location of the key file |
 | -listen         | :9710  | address to listen for incoming requests  |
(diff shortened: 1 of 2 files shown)


secustor (changes requested):
This change should be also reflected in the `Readme.md`. 

Also this commit type should be `feat`

bluecmd:
> Also this commit type should be feat

I don't agree, but I will change it. My argument is that this doesn't change anything anyone could reasonably depend on (the default value right now is invalid and only causes crashes).

secustor:
I agree with you, if somebody could depend on it, it should be a breaking change. 
But `minor` will not generate a release and as I understand you that is your intend.

secustor (approved):

bluecmd:
Ah, right. You are correct :-)

bluecmd:
What about `fix(flag)` instead? It feels a bit too much to bump to `1.8.0` just because of this.

PR #111, opened 2021-07-10
refactor(ci): Remove setup-env phase

bluecmd:
setup-env relies on that go install can install to $PATH. If users have
not configured their environment, using go run is easier on them.


bluecmd:
I have to check if `go run` outputs extra stuff on the first run, that might be a problem.

secustor (approved):

PR #115, opened 2021-07-12
Add BGP neighbour path and best path count

bl0m1:
I do not know if this metric exists before 7.0.0. Need to investigate this, if anyone runs a FW on 6.X it would be super nice if you can verify the path and path6 endpoint.


README.md
```diff
@@ -95,6 +93,14 @@ Per-VDOM:
  * `fortigate_virtual_wan_bandwidth_rx_byte_per_second`
  * `fortigate_virtual_wan_status_change_time_seconds`
 
+ Per-BGP-Neighbor and VDOM:
+ * `fortigate_bgp_neighbor_ipv4_info`
+ * `fortigate_bgp_neighbor_ipv6_info`
+ * `fortigate_bgp_neighbor_paths_ipv4_count`
```
bluecmd, line 99: The `_count` suffix is sort of reserved for Prometheus counters, and this is a  Gauge. So maybe call these `fortigate_bgp_neighbor_ipv4_paths` and `fortigate_bgp_neighbor_ipv4_best_paths`?

pkg/probe/bgp_neighbor_routes.go
```diff
@@ -0,0 +1,161 @@
...
+	Version string    `json:"version"`
+}
+
+type PathCount struct {
+	Source string
+	VDOM   string
+}
+
+func probeBGPNeighborPathsIPv4(c http.FortiHTTP, meta *TargetMetadata) ([]prometheus.Metric, bool) {
+	MaxBGPPaths, MaxBGPPathsSet := os.LookupEnv("FORTIGATE_EXPORTER_MAX_BGP_PATHS")
```
bluecmd, line 30: Use a flag.Int instead since that is what we are using today.

pkg/probe/bgp_neighbor_routes.go
```diff
@@ -0,0 +1,161 @@
...
+		BGPNeighborBestPaths = prometheus.NewDesc(
+			"fortigate_bgp_neighbor_best_paths_ipv4_count",
+			"Count of best paths for an BGP neighbor",
+			[]string{"vdom", "neighbor_ip"}, nil,
+		)
+	)
+
+	var rs []BGPPaths
+
+	if err := c.Get("api/v2/monitor/router/bgp/paths", fmt.Sprintf("vdom=*&count=%s", MaxBGPPaths), &rs); err != nil {
```
bluecmd, line 54: For numbers use `[..]=%d`

pkg/probe/bgp_neighbor_routes.go
```diff
@@ -0,0 +1,161 @@
...
+
+	m := []prometheus.Metric{}
+
+	srMap := make(map[PathCount]int)
+	sr2Map := make(map[PathCount]int)
+	for _, r := range rs {
+
+		iMaxBGPPaths, _ := strconv.Atoi(MaxBGPPaths)
+		if len(r.Results) >= iMaxBGPPaths {
+			log.Printf("Error: Recived more BGP Paths than maximum allowed, ignoring metric ...")
```
bluecmd, line 67:
```suggestion
			log.Printf("Error: Received more BGP Paths than maximum (%d > %d) allowed, ignoring metric ...", len(r.Results), iMaxBGPPaths)
```

pkg/probe/bgp_neighbor_routes_test.go
```diff
@@ -0,0 +1,57 @@
...
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/testutil"
+)
+
+func TestBGPNeighborPathsIPv4(t *testing.T) {
```
bluecmd, line 11: Add:
```go
	if err := config.Init(); err != nil {
		t.Fatalf("config.Init failed: %+v", err)
	}
```

pkg/probe/bgp_neighbor_routes.go
```diff
@@ -0,0 +1,154 @@
...
+				sr2 := PathCount{
+					Source: route.LearnedFrom,
+					VDOM:   r.VDOM,
+				}
+				sr2Map[sr2] += 1
+			}
+		}
+	}
+
+	for neighbor, i := range srMap {
```
secustor, line 154: Please replace `i` with something more speaking such like `count`. `i` suggests that it is an index rather a value.

pkg/probe/bgp_neighbor_routes.go
```diff
@@ -0,0 +1,154 @@
...
+		m = append(m, prometheus.MustNewConstMetric(BGPNeighborPaths, prometheus.GaugeValue, float64(i), neighbor.VDOM, neighbor.Source))
+	}
+	for neighbor, i := range sr2Map {
```
secustor, line 157: same here

pkg/probe/bgp_neighbor_routes.go
```diff
@@ -0,0 +1,154 @@
...
+				sr2 := PathCount{
+					Source: route.LearnedFrom,
+					VDOM:   r.VDOM,
+				}
+				sr2Map[sr2] += 1
+			}
+		}
+	}
+
+	for neighbor, i := range srMap {
```
secustor, line 86: Please replace i with something more speaking such like count. i suggests that it is an index rather a value.

pkg/probe/bgp_neighbor_routes.go
```diff
@@ -0,0 +1,154 @@
...
+		m = append(m, prometheus.MustNewConstMetric(BGPNeighborPaths, prometheus.GaugeValue, float64(i), neighbor.VDOM, neighbor.Source))
+	}
+	for neighbor, i := range sr2Map {
```
secustor, line 89: same here

pkg/probe/bgp_neighbor_routes_test.go
```diff
@@ -0,0 +1,57 @@
...
+	c.prepare("api/v2/monitor/router/bgp/paths", "testdata/router-bgp-paths-v4.jsonnet")
+	r := prometheus.NewPedanticRegistry()
+	if !testProbe(probeBGPNeighborPathsIPv4, c, r) {
+		t.Errorf("probeBGPNeighborPathsIPv4() returned non-success")
+	}
+
+	em := `
+    # HELP fortigate_bgp_neighbor_best_paths_ipv4_count Count of best paths for an BGP neighbor
+    # TYPE fortigate_bgp_neighbor_best_paths_ipv4_count gauge
+    fortigate_bgp_neighbor_best_paths_ipv4_count{neighbor_ip="10.0.0.1",vdom="root"} 1
```
secustor, line 29: Maybe rename this more in line of `fortigate_bgp_neighbor_propagated_paths_ipvX` and `fortigate_bgp_neighbor_propagated_best_paths_ipvX`? @bl0m1 @bluecmd

secustor (changes requested):

pkg/probe/bgp_neighbor_routes_test.go
bl0m1: I'm open for this, The current naming is just a copy paste of the API endpoint and value naming.

pkg/probe/bgp_neighbor_routes_test.go
bl0m1: I got this working with your fix :) I also added a test "config" file to avoid getting errors when testing :)

pkg/probe/fortigate-key.yaml
```diff
@@ -0,0 +1,2 @@
+"127.0.0.1":
```
bluecmd, line 1:
```suggestion
# This file is used for running test cases
"127.0.0.1":
```

pkg/probe/bgp_neighbor_routes_test.go
bluecmd: All paths are `propagated`, the `best_paths` are just the ones that are active / in-use. Since the algorithm for selecting these are called "BGP Best Path Selection" I think it makes sense to stick to the terminology.

bluecmd (approved):

secustor (approved):

Issue #64, opened 2021-03-14
Add Contributing.md


secustor:
To pull together all information regarding adding code to this project, we should add a `Contributing.md`. 

The content should contain at least following sections: 
 - How to build this project
 - metric best practices
 - links to resources 
 - allowed commit message styles and scopes

secustor:
reference: https://github.com/bluecmd/fortigate_exporter/issues/100#issuecomment-877096726

IMSdevsecu:
Now addressed in #141

PR #118, opened 2021-08-04
Add probe selection capability from configuration file

IMSdevsecu:
Hi,

First, thank you to all contributors for this great job. This exporter is helping us a lot every day.

I propose a capability to select only specific probes instead of all by default.

Objectives are:
- optimize scraping to only needed/interesting probes in order to save network and fortigate load
- optimize scraping at a large scale, when scraping large number of fortigate
- offer a solution to handle probing specificities for each fortios version in configuration file rather than in code  
- prevent continuous errors in logs about some probes failing

For example, I am getting a 403 for api/v2/monitor/firewall/load-balance on multiple fortios 6.0.10.

The code I propose use the configuration file to list probes to run for each fortigate.

There is 2 ways to add probes:
- by their name: add probes one by one
- by their category: add a group of probes grouped by category (ex: BGP, System ...)

This is an option being added, therefore there is no issue with retro compatibility.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,137 +29,160 @@ Supported metrics right now as follows.
 
 Global:
 
- * `fortigate_version_info`
- * `fortigate_cpu_usage_ratio`
- * `fortigate_memory_usage_ratio`
- * `fortigate_current_sessions`
- * `fortigate_license_vdom_usage`
- * `fortigate_license_vdom_max`
+ * _System/Status_
+   * `fortigate_version_info`
+ * _System/Resource/Usage_
+   * `fortigate_cpu_usage_ratio`
+   * `fortigate_memory_usage_ratio`
+   * `fortigate_current_sessions`
+ * _License/Status_
+   * `fortigate_license_vdom_usage`
+   * `fortigate_license_vdom_max`
 
 Per-VDOM:
 
- * `fortigate_vdom_cpu_usage_ratio`
- * `fortigate_vdom_memory_usage_ratio`
- * `fortigate_vdom_current_sessions`
- * `fortigate_policy_active_sessions`
- * `fortigate_policy_bytes_total`
- * `fortigate_policy_hit_count_total`
- * `fortigate_policy_packets_total`
- * `fortigate_interface_link_up`
- * `fortigate_interface_speed_bps`
- * `fortigate_interface_transmit_packets_total`
- * `fortigate_interface_receive_packets_total`
- * `fortigate_interface_transmit_bytes_total`
- * `fortigate_interface_receive_bytes_total`
- * `fortigate_interface_transmit_errors_total`
- * `fortigate_interface_receive_errors_total`
- * `fortigate_vpn_connections`
- * `fortigate_ipsec_tunnel_receive_bytes_total`
- * `fortigate_ipsec_tunnel_transmit_bytes_total`
- * `fortigate_ipsec_tunnel_up`
- * `fortigate_wifi_access_points`
- * `fortigate_wifi_fabric_clients`
- * `fortigate_wifi_fabric_max_allowed_clients`
+ * _System/VDOMResources_
+   * `fortigate_vdom_cpu_usage_ratio`
+   * `fortigate_vdom_memory_usage_ratio`
+   * `fortigate_vdom_current_sessions`
+ * _Firewall/Policies_
+   * `fortigate_policy_active_sessions`
+   * `fortigate_policy_bytes_total`
+   * `fortigate_policy_hit_count_total`
+   * `fortigate_policy_packets_total`
+ * _System/Interface_
+   * `fortigate_interface_link_up`
+   * `fortigate_interface_speed_bps`
+   * `fortigate_interface_transmit_packets_total`
+   * `fortigate_interface_receive_packets_total`
+   * `fortigate_interface_transmit_bytes_total`
+   * `fortigate_interface_receive_bytes_total`
+   * `fortigate_interface_transmit_errors_total`
+   * `fortigate_interface_receive_errors_total`
+ * _VPN/Ssl_
+   * `fortigate_vpn_connections`
+ * _VPN/IPSec_
+   * `fortigate_ipsec_tunnel_receive_bytes_total`
+   * `fortigate_ipsec_tunnel_transmit_bytes_total`
+   * `fortigate_ipsec_tunnel_up`
+ * _Wifi/APStatus_
+   * `fortigate_wifi_access_points`
+   * `fortigate_wifi_fabric_clients`
+   * `fortigate_wifi_fabric_max_allowed_clients`
 
  Per-HA-Member and VDOM:
- * `fortigate_ha_member_info`
- * `fortigate_ha_member_cpu_usage_ratio`
- * `fortigate_ha_member_memory_usage_ratio`
- * `fortigate_ha_member_network_usage_ratio`
- * `fortigate_ha_member_sessions`
- * `fortigate_ha_member_packets_total`
- * `fortigate_ha_member_virus_events_total`
- * `fortigate_ha_member_bytes_total`
- * `fortigate_ha_member_ips_events_total`
+ * _System/HAStatistics_
+   * `fortigate_ha_member_info`
+   * `fortigate_ha_member_cpu_usage_ratio`
+   * `fortigate_ha_member_memory_usage_ratio`
+   * `fortigate_ha_member_network_usage_ratio`
+   * `fortigate_ha_member_sessions`
+   * `fortigate_ha_member_packets_total`
+   * `fortigate_ha_member_virus_events_total`
+   * `fortigate_ha_member_bytes_total`
+   * `fortigate_ha_member_ips_events_total`
 
- Per-Link and VDOM: 
- * `fortigate_link_status`
- * `fortigate_link_latency_seconds`
- * `fortigate_link_latency_jitter_seconds`
- * `fortigate_link_packet_loss_ratio`
- * `fortigate_link_packet_sent_total`
- * `fortigate_link_packet_received_total`
- * `fortigate_link_active_sessions`
- * `fortigate_link_bandwidth_tx_byte_per_second`
- * `fortigate_link_bandwidth_rx_byte_per_second`
- * `fortigate_link_status_change_time_seconds`
+ Per-Link and VDOM:
+ * _System/LinkMonitor_
+   * `fortigate_link_status`
+   * `fortigate_link_latency_seconds`
+   * `fortigate_link_latency_jitter_seconds`
+   * `fortigate_link_packet_loss_ratio`
+   * `fortigate_link_packet_sent_total`
+   * `fortigate_link_packet_received_total`
+   * `fortigate_link_active_sessions`
+   * `fortigate_link_bandwidth_tx_byte_per_second`
+   * `fortigate_link_bandwidth_rx_byte_per_second`
+   * `fortigate_link_status_change_time_seconds`
 
  Per-SDWAN and VDOM:
- * `fortigate_virtual_wan_status`
- * `fortigate_virtual_wan_latency_seconds`
- * `fortigate_virtual_wan_latency_jitter_seconds`
- * `fortigate_virtual_wan_packet_loss_ratio`
- * `fortigate_virtual_wan_packet_sent_total`
- * `fortigate_virtual_wan_packet_received_total`
- * `fortigate_virtual_wan_active_sessions`
- * `fortigate_virtual_wan_bandwidth_tx_byte_per_second`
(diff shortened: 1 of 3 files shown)


bluecmd:
Thanks for the kind words. I think this looks fantastic! I have no objections to merge this as is, what do you say @secustor ?

README.md
```diff
@@ -183,6 +183,22 @@ Where `fortigate-key.yaml` contains pairs of FortiGate targets and API keys in t
 NOTE: Currently only token authentication is supported. FortiGate does not allow usage of tokens on non-HTTPS connections,
 which means that currently you need HTTPS to be configured properly.
 
+You can select which probes or probe categories you want to run per target, for example:
+
+```
+"https://my-fortigate":
+  token: api-key-goes-here
+  probes:
+    - System
```
bluecmd, line 192: Since it is not clear if `System` is a probe or a category, maybe we should do some kind of prefix tier system?

Like you define your metric as `/BGP/BGPNeighborsIPv4` - and that means that somebody can do either:
```yaml
probes:
  - /BGP
```
or
```yaml
probes:
  - /BGP/BGPNeighborsIPv4
```

Code-wise we could just use something like `strings.HasPrefix`.

IMSdevsecu: I am OK with your tier modification.
This is actually more convenient and flexible on long term with an increasing number of probes.
I think the best solution is to support both :
- probe unique full name (ex: `BGPNeighborsIPv4`) - must be equal to current `aProbe.name`
- probe unique category full name (ex: `BGP/Neighbors/IPv4`) - must `HasPrefix` with `aProbe.category`

Then `BGP/Neighbors/IPv4` and `BGPNeighborsIPv4` are equivalent and can both be used to identify a unique probe.

For prefix matching, HasPrefix offers flexible solution for, multi tier match at any level.
Example :
```
- BGP                   (match all BGP probes)
- BGP/                  (match all BGP probes)
- BGP/Neighb            (match all 4 BGP probes)
- BGP/Neighbors         (match 2 BGP probes BGPNeighborsIPv4 and BGPNeighborsIPv6)
- BGP/Neighbors/        (match 2 BGP probes BGPNeighborsIPv4 and BGPNeighborsIPv6)
- BGP/Neighbors/IPv     (match 2 BGP probes BGPNeighborsIPv4 and BGPNeighborsIPv6)
- BGP/Neighbors/IPv4    (match 1 BGP probes BGPNeighborsIPv4)
- BGP/Neighbors/IPv44   (match nothing)
```
This solution looks quite optimized and polyvalent .


One step forward, we can obtain `BGPNeightborsIPv4` from `BGP/Neighbors/IPv4` by just removing slashes.
This simplifies the current struct to:
```golang
type probeDetailedFunc struct {
        name string
	function probeFunc
}
````
`name` is now storing former category (ex: `BGP/Neighbors/IPv4`).
`probe.go` declaration section would be more simple.

Do you think this is a good idea to go with the slashes removing, instead of storing both category and name?

Comment (secustor) [reply README.md:192]: I really like the idea of this prefix system, but I propose a slight modification. 

### All probes are executed
``` yaml
"https://my-fortigate":
  token: api-key-goes-here
    # or entirely dropped -> all probes are used
``` 
``` yaml
"https://my-fortigate":
  token: api-key-goes-here
  probes:  # or entirely dropped -> all probes are used
``` 
``` yaml
"https://my-fortigate":
  token: api-key-goes-here
  probes:
    include: [ ] # -> all probes are used
```
### Only matching probes are executed
``` yaml
"https://my-fortigate":
  token: api-key-goes-here
  probes:
    include: 
      - System
```


This allows to add additional options later on such like this and is more readable: 
``` yaml
"https://my-fortigate":
  token: api-key-goes-here
  probes:
    include: 
      - System
    exclude: 
      - System/SystemHAStatistics
```


Regarding the slash replacement.    
I'm not sure what you mean with 'slash removing' as I understand your latest proposal would be to store in the style of : 
```
[
  { "System/SystemHAStatistics", probeSystemHAStatistics },
...
]
```

secustor (changes requested):
Please add the categories and how to use them to the README, when we have settled on a solution

README.md
IMSdevsecu: By 'slash removing' I mean we can just store in `probe.go`
```
for _, aProbe := range []probeDetailedFunc{
		{"BGP/Neighbors/IPv4", probeBGPNeighborsIPv4},
                {"System/HAStatistics", probeSystemHAStatistics},
                ...
```
User can either add `BGP/Neighbors/IPv4` or `BGPNeighborsIPv4` which is totally equivalent for including this probe.

### Exclusion with "!"
I was also thinking to an exclude capability as a next step, but as you talked about it we can plan to add it all at once.
Your proposal looks fine to me.

I was more thinking to a "!" syntax.
Processing of inclusion, exclusion could be done sequentially.
The logic :
- if list start with an include statement, then initialize all other probes to "not wanted"
- if list start with an exclude statement, then initialize all probes to "wanted", expect those matching the exclude statement
- for 2nd to N term of the list, just include or exclude matching probes for each list entry

Summing up and detailing use cases :
```yaml
probes: System
# select only all System prefixed probes

probes: BGP/Neighbors/IPv4
# select only BGP/Neighbors/IPv4

probes: BGPNeighborsIPv4
# equivalent to previous declaration

probes: BGPNeighborsIPv4, BGP/Neighbors/IPv6
# mixing inclusion system category/name to include both BGPNeighborsIPv4 AND BGPNeighborsIPv6

probes: System, BGP
# select all System AND BGP prefixed probes

probes: !System
# select all probes but those prefixed by System

probes: !System, BGP
# not really clever but select all probes but those prefixed by System

probes: BGP, !System
# not really clever but just select all BGP prefixed probes

probes: !BGP, BGP/Neighbors
# select all probes but those prefixed by BGP + the 2 probes starting with BGP/Neighbors

probes: BGP/Neighbors, !BGP
# actually select no probe, because first select 2 BGP probes, excluded by !BGP

probes: BGP/Neighbors, !BGP, BGP/Neighbors
# not really clever but just select the 2 probes starting with BGP/Neighbors, then exclude all BGP probes, and finally select the 2 probes starting with BGP/Neighbors again
```

pros :
- one list declaration for include/exclude
- possible do declare a complex inclusion exclusion sequence (order matters)
- go through 1 list instead of 2

cons :
- require to add a check on "!" prefix for probe name
- maybe a bit less simple to read

What do you think about this inclusion/exclusion method ?

### Probe name constraints
I don't know if there is already a naming convention for probes.
I propose the following probe name constraints to make probe naming more strait forward from Fortigate API.
- full name ex `SystemHAstatistics` is obtained from categorized name ex: `System/HAstatistics` which is stored in `pkg/probe/probe.go`
- probe name cannot start with "!"
- probe full name should respect regex : `^[a-ZA-Z0-9]{4,}$`
- probe categorized name must respect regex : `^([a-ZA-Z0-9]+/)+[a-ZA-Z0-9]+$`
- probe name should follow the fortigate API syntax ex: `System/HAstatistics` for `/api/v2/monitor/system/ha-statistics`
- probe full name and categorized name must be unique across all probes

Right now, not all probes respect the entire naming convention, for exemple:
`/api/v2/monitor/router/bgp/neighbors` should be `router/BGP/Neighbors/IPv4` and `routerBGPNeighborsIPv4`

What do you think of those probe name constraints ? Do you think applying this to the entire project is the right thing to do ?

bluecmd: I am with @secustor here - I like include/exclude better than `!` for the simple fact that it is less string-mangling. We have a structured configuration file format already (YAML), we are probably better off using it to communicate meaning as much as possible.

I am a fan of having less ways of doing the same thing, so if we can agree on that the probe name is for example `BGP/Neighbors/IPv4` then I would prefer that be the only way it would be written. I think introducing multiple ways of referring to a single probe adds a lot of confusion for a small benefit for some people that do not like the `/` character.

Finally - I am all for your naming rules @IMSdevsecu, we should add them to the contribution guide (which isn't written yet, https://github.com/bluecmd/fortigate_exporter/issues/64). The only part I want to relax is the "should follow Fortigate API" bit. Why? Many metrics requires fetting both from monitor and potentially cmdb which both [can vary](https://github.com/bluecmd/fortigate_exporter/blob/e4274022852cfcc6f196180bf88fa3ba3d57f11d/pkg/probe/system_ha_statistics.go#L86-L98) in how they look like - besides, we probably don't want to break people's configs just because Fortinet changed their APIs between e.g. 6.x and 7.x.

Instead, I would formulate your "probe name should follow the fortigate API syntax" as a recommendation as to what to name the probe function when it is first introduced, with two notable exceptions:
- If there already exists similar metrics, stick to that naming convention
- Shorter is better (E.g. `Router/BGP/x` is redundant, use `BGP/x`)

Also, @secustor - another example that follows logically given that we are prefix matching:
```yaml
"https://my-fortigate":
  token: api-key-goes-here
  probes:
    include: 
       - '' # -> all probes are used
```

IMSdevsecu: > I am with @secustor here - I like include/exclude better than `!` for the simple fact that it is less string-mangling. We have a structured configuration file format already (YAML), we are probably better off using it to communicate meaning as much as possible.

I understand. This is more simple to code, use and document. This also allows for less complex selection.
If we go for 2 distinct lists which one to process first ?
Because default is "all probe wanted", I think we can process this way:
```
foreach probe 
    if "include" list is not empty then
        default to "not wanted"
        if probe as Prefix in "include" list then
            set to "wanted"
    else
        default to "wanted"

    if "exclude" list is not empty then
         if probe as Prefix in "exclude" list then
            set to "not wanted"
```
Is this algorithm okay for you ?

IMSdevsecu: > I am a fan of having less ways of doing the same thing, so if we can agree on that the probe name is for example `BGP/Neighbors/IPv4` then I would prefer that be the only way it would be written. I think introducing multiple ways of referring to a single probe adds a lot of confusion for a small benefit for some people that do not like the `/` character.

Fine let's keep the `BGP/Neighbors/IPv4` name only.

IMSdevsecu: > Finally - I am all for your naming rules @IMSdevsecu, we should add them to the contribution guide (which isn't written yet, #64). The only part I want to relax is the "should follow Fortigate API" bit. Why? Many metrics requires fetting both from monitor and potentially cmdb which both [can vary](https://github.com/bluecmd/fortigate_exporter/blob/e4274022852cfcc6f196180bf88fa3ba3d57f11d/pkg/probe/system_ha_statistics.go#L86-L98) in how they look like - besides, we probably don't want to break people's configs just because Fortinet changed their APIs between e.g. 6.x and 7.x.
> 
> Instead, I would formulate your "probe name should follow the fortigate API syntax" as a recommendation as to what to name the probe function when it is first introduced, with two notable exceptions:
> 
>     * If there already exists similar metrics, stick to that naming convention
> 
>     * Shorter is better (E.g. `Router/BGP/x` is redundant, use `BGP/x`)
> 
> 
> Also, @secustor - another example that follows logically given that we are prefix matching:
> 
> ```yaml
> "https://my-fortigate":
>   token: api-key-goes-here
>   probes:
>     include: 
>        - '' # -> all probes are used
> ```

Sounds good to me.
You must have a deeper knowledge of the evolution of their API. I trust you if you prefer to keep some flexibility with the naming 😉

README.md
```diff
@@ -183,21 +206,53 @@ Where `fortigate-key.yaml` contains pairs of FortiGate targets and API keys in t
...
 ```
 "https://my-fortigate":
   token: api-key-goes-here
   probes:
-    - System
+    include:
+      - System
+      - VPN
+      - Firewall/Policies
+      # Inlcude only probes with name starting with: System or VPN + probe: Firewall/Policies
```
bluecmd, line 227:
```suggestion
      # Include only probes with name starting with: System or VPN + probe: Firewall/Policies
```

README.md
```diff
@@ -183,21 +206,53 @@ Where `fortigate-key.yaml` contains pairs of FortiGate targets and API keys in t
...
+      # Inlcude probes with name starting with: System and Firewall
+      # Then exclude probe: System/LinkMonitor
 ```
 
-If `probes` isn't set or is empty, all probes will be run against the target.
+Special cases:
+
+- If `probes` isn't set or is empty, all probes will be run against the target.
+- If `include` list is empty, by default, all probes will be selected to be run against the target.
+- If `include` contains an entry `- ''`, then all probes are inlcuded (equivalent to not defining `include`)
```
bluecmd, line 253:
```suggestion
- If `include` contains an entry `- ''`, then all probes are included (equivalent to not defining `include`)
```

bluecmd (approved):

bluecmd:
I think this looks really neat. What do you say @secustor ?

secustor (approved):
LGTM thanks for your contribution and the exhaustive documentation!

PR #119, opened 2021-08-04
feat(probe): Add probe for vpn SSL users

IMSdevsecu:
Hi,

We are interested in listing connected user to SSL VPN for example in Grafana dashboards.
This code add support for probing SSL VPN users in addition to other metrics.

The code also include a new parameter to limit number of retrieved users.

### DIFF
--- a/internal/config/main.go
+++ b/internal/config/main.go
@@ -61,7 +61,7 @@ var (
 		TLSInsecure:   flag.Bool("insecure", false, "Allow insecure certificates"),
 		TlsExtraCAs:   flag.String("extra-ca-certs", "", "comma-separated files containing extra PEMs to trust for TLS connections in addition to the system trust store"),
 		MaxBGPPaths:   flag.Int("max-bgp-paths", 10000, "How many BGP Paths to receive when counting routes, needs to be higher then the number of routes or metrics will not be generated"),
-		MaxVPNUsers:   flag.Int("max-vpn-users", 10000, "How many VPN Users to receive when counting users, needs to be greater than or equal the number of users or metrics will not be generated"),
+		MaxVPNUsers:   flag.Int("max-vpn-users", 0, "How many VPN Users to receive when counting users, needs to be greater than or equal the number of users or metrics will not be generated (0 eq. none by default)"),
 	}
 
 	savedConfig *FortiExporterConfig
--- a/internal/config/main.go
+++ b/internal/config/main.go
@@ -17,6 +17,7 @@ type FortiExporterParameter struct {
 	TLSInsecure   *bool
 	TlsExtraCAs   *string
 	MaxBGPPaths   *int
+	MaxVPNUsers   *int
 }
 
 type FortiExporterConfig struct {
@@ -27,6 +28,7 @@ type FortiExporterConfig struct {
 	TLSInsecure   bool
 	TlsExtraCAs   []LocalCert
 	MaxBGPPaths   int
+	MaxVPNUsers   int
 }
 
 type AuthKeys map[Target]TargetAuth
@@ -59,6 +61,7 @@ var (
 		TLSInsecure:   flag.Bool("insecure", false, "Allow insecure certificates"),
 		TlsExtraCAs:   flag.String("extra-ca-certs", "", "comma-separated files containing extra PEMs to trust for TLS connections in addition to the system trust store"),
 		MaxBGPPaths:   flag.Int("max-bgp-paths", 10000, "How many BGP Paths to receive when counting routes, needs to be higher then the number of routes or metrics will not be generated"),
+		MaxVPNUsers:   flag.Int("max-vpn-users", 10000, "How many VPN Users to receive when counting users, needs to be greater than or equal the number of users or metrics will not be generated"),
 	}
 
 	savedConfig *FortiExporterConfig
@@ -86,6 +89,7 @@ func ReInit() error {
 		TLSTimeout:    *parameter.TLSTimeout,
 		TLSInsecure:   *parameter.TLSInsecure,
 		MaxBGPPaths:   *parameter.MaxBGPPaths,
+		MaxVPNUsers:   *parameter.MaxVPNUsers,
 	}
 
 	// parse AuthKeys
--- a/pkg/probe/vpn_ssl.go
+++ b/pkg/probe/vpn_ssl.go
@@ -3,33 +3,62 @@ package probe
 import (
 	"log"
 
+	"github.com/bluecmd/fortigate_exporter/internal/config"
 	"github.com/bluecmd/fortigate_exporter/pkg/http"
 	"github.com/prometheus/client_golang/prometheus"
 )
 
+type VPNUser struct {
+	UserName string `json:"user_name"`
+}
+
+type VPNUsers struct {
+	Results []VPNUser `json:"results"`
+	VDOM    string    `json:"vdom"`
+	Version string    `json:"version"`
+}
+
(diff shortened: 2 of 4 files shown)


internal/config/main.go
```diff
@@ -52,6 +54,7 @@ var (
 		TLSInsecure:   flag.Bool("insecure", false, "Allow insecure certificates"),
 		TlsExtraCAs:   flag.String("extra-ca-certs", "", "comma-separated files containing extra PEMs to trust for TLS connections in addition to the system trust store"),
 		MaxBGPPaths:   flag.Int("max-bgp-paths", 10000, "How many BGP Paths to receive when counting routes, needs to be higher then the number of routes or metrics will not be generated"),
+		MaxVPNUsers:   flag.Int("max-vpn-users", 10000, "How many VPN Users to receive when counting users, needs to be greater than or equal the number of users or metrics will not be generated"),
```
bluecmd, line 57: This makes a few warning bells go off in my head. I am thinking two issues:

1. The sheer amount of metrics this will create is likely quite huge given it is per user. It is configurable for the maximum number of users though, so that's OK.
2. If the iteration order changes, so does the potential order of the output - which is fine, given that we emit no partial metrics (this would be a problem if we e.g. emitted *up to* 1000 first users and capped it at that).

I think this will work, and I can see why people would like to have this metric available. It scares me though.

Could we lower the default limit to 100 users maybe? For the BGP case we do not (unless I remember incorrectly?) emit a metric for every path we fetch. Here the very likely case is that every user will have a small amount of sessions, so I think limiting it to a maximum of 100 metrics would be a sensible default.

IMSdevsecu: >     1. The sheer amount of metrics this will create is likely quite huge given it is per user. It is configurable for the maximum number of users though, so that's OK.

This can for sure create a lot of metrics. Cardinality is always an issue for this kind of metrics.

IMSdevsecu: >     2. If the iteration order changes, so does the potential order of the output - which is fine, given that we emit no partial metrics (this would be a problem if we e.g. emitted _up to_ 1000 first users and capped it at that).

Yes, there should be no partial metrics; therefore we can expect consistency across scrapes.

IMSdevsecu: > I think this will work, and I can see why people would like to have this metric available. It scares me though.

My original aim is to be able to trigger an alert if a user is connected with SSL VPN in certain conditions.
For example:
-	User is connect while he is not supposed to. This could be acting as a double check for highly critical access where fortigate configuration is already pretty restricted.
-	User is connect at some hours it should not. IT team member of the company for example have emergency access but should use it only if required.

For our needs, this will only be used for detecting abnormal security behavior.
No aim to spy on users activity, but I am not responsible for people motivations outside of security.
We can just advise people not to be evil.

IMSdevsecu: > Could we lower the default limit to 100 users maybe? For the BGP case we do not (unless I remember incorrectly?) emit a metric for every path we fetch. Here the very likely case is that every user will have a small amount of sessions, so I think limiting it to a maximum of 100 metrics would be a sensible default.

Ok for a default limit of 100. I guess this fit most deployment scenarios.
To force people not to be evil we can also lower it to 0 by default and add a check to not run the API call if set to 0.
Therefore, only users that want to retrieve this metric would have to set the parameter to a higher value they will have to choose.
What do you think of disabling this metric by default by setting the limit to 0 ?

IMSdevsecu: @bluecmd @secustor Do you think disabling this probe by default by setting it to 0 is fine? Let me know and I will update code accordingly

bluecmd: I am OK with setting the value to 0 or a small value (<100) by default. Either way works for me.

Can you rebase (not merge) on the latest? It seems there are conflicts and the diffs are all messed up sadly.

IMSdevsecu: Sorry for the mess. I am not very confident with rebase, I never used it before. Hope this is better now. If not tell me, and I will try harder.

bluecmd: It's fine, it's one of the harder things to do in git.

You want to do:
```
$ git fetch origin
$ git rebase -i origin/master
$ git push --force-with-lease
```

When you do `git rebase -i` you will see the commits that it wants to bring over "ontop" of the new `master`. Those commits should be the ones that contain your modifications, nothing else.

IMSdevsecu:
Thank you for the git rebase tip, not sure they are perfectly ordered, but will do better next time

pkg/probe/vpn_ssl.go
```diff
@@ -3,33 +3,59 @@ package probe
...
+}
+
+type VPNUsers struct {
+	Results []VPNUser `json:"results"`
+	VDOM    string    `json:"vdom"`
+	Version string    `json:"version"`
+}
+
 func probeVPNSsl(c http.FortiHTTP, meta *TargetMetadata) ([]prometheus.Metric, bool) {
+	savedConfig := config.GetConfig()
```
bluecmd, line 22: I don't like this very much, but I understand why you went this path.

I sent out https://github.com/bluecmd/fortigate_exporter/pull/125 for another way that I like more.
This would cause this part to become:
```go
import "flag"
...
        flag.Set("max-vpn-users", "10")
        config.MustReInit()
        if !testProbe(probeVPNSsl, c, r) {
                t.Errorf("probeVPNSsl() returned non-success")
        }
```

bluecmd: Ready for your usage now :)

IMSdevsecu (commented):
@bluecmd Your way is much better :) code updated

bluecmd (approved):
LGTM, what do you say @secustor ?

pkg/probe/vpn_ssl_test.go
```diff
@@ -12,6 +14,8 @@ func TestVPNSsl(t *testing.T) {
 	c := newFakeClient()
 	c.prepare("api/v2/monitor/vpn/ssl", "testdata/vpn.jsonnet")
 	r := prometheus.NewPedanticRegistry()
+	flag.Set("max-vpn-users", "10")
```
secustor, line 17:
```suggestion
	flag.Set("max-vpn-users", "0")
```

Should be 0, if reading the discussions.

internal/config/main.go
```diff
@@ -59,6 +61,7 @@ var (
 		TLSInsecure:   flag.Bool("insecure", false, "Allow insecure certificates"),
 		TlsExtraCAs:   flag.String("extra-ca-certs", "", "comma-separated files containing extra PEMs to trust for TLS connections in addition to the system trust store"),
 		MaxBGPPaths:   flag.Int("max-bgp-paths", 10000, "How many BGP Paths to receive when counting routes, needs to be higher then the number of routes or metrics will not be generated"),
+		MaxVPNUsers:   flag.Int("max-vpn-users", 0, "How many VPN Users to receive when counting users, needs to be greater than or equal the number of users or metrics will not be generated (0 eq. none by default)"),
```
secustor, line 64: What is the rational behind a complete abort if max-vpn-users is smaller than the results?     
 As user I would expect that we simply stop adding new metrics.

secustor (changes requested):

pkg/probe/vpn_ssl_test.go
secustor: Ignore this I have overlooked that this is a test

internal/config/main.go
bluecmd: The problem is that if we do that, we assume that the order of all entries are the same for every API request from the fortigate, or we would end up getting partial results that differ for every scrape.

This may hold today (no idea if it does), but may change anytime in the future.

secustor (approved):

PR #127, opened 2021-08-19
fix(probe): VPN/Ssl users: Fix multiple vpn for a single username

IMSdevsecu:
Hi,

There is an issue while a single username is used to setup multiple SSL VPN.
Unicity is not satified with only `vdom` and `user` as labels.

Here is the error returned:
```
collected metric "fortigate_vpn_users" { label:<name:"user" value:"user1" > label:<name:"vdom" value:"root" > gauge:<value:1 > } was collected before with the same name and label values
```

I added some more tests and added 2 labels to have unicity: `rhost` and `aip`


IMSdevsecu:
This is possible. My approach was, what can do more can do less :)
I also thouth this was the best place to show up those extra labels using the matric `fortigate_vpn_users`.

If we go for a counter of number of connections per username, then we start sumurarizing, and we will have to setup another metric to show up IP if we want to.

I was also thinking to show `in_bytes` `out_bytes` as value, to maximize information carried and level of detail instead of simply 1.
For example we can show two metrics `fortigate_vpn_users_receive_bytes_total` and  `fortigate_vpn_users_transmit_bytes_total` instead of `fortigate_vpn_users` but it is maybe to late.
There will not be that much to expose on that section of the API then.
I know it would have been probably better to start this way, but I have not though about 1 user / 2 connections at first and it did not append while running my tests.
I suggest this possibility as this new metric it is still fresh and not activated by default, then a lot of people will propably not miss it :)
@bluecmd @secustor  What do you think of replacing the current metric by those 2 metrics ?

Anyway, having a second look to the fortigate API description, this may be a good think to expose an index to have maximum uniqueness guarantees.
@bluecmd @secustor What do you think of exposing both index, i.e. the one in the subsession block and the one out of it?
Only index in subsession could be enough for unicity (looks like it) but I am not sure.
I also hope index are not renumbered during a connection, this could be messy...

bluecmd:
Some quick comments

> @bluecmd @secustor What do you think of replacing the current metric by those 2 metrics ?

I don't think we should remove any metrics that have been added unless we are bumping the major version to 2.0. People rely on stability of metrics when it comes to exporters, it is part of our API. Even if it was very recently added.

Adding new metrics is totally fine though, of course.

> I also hope index are not renumbered during a connection, this could be messy...

I agree, and I don't think we should rely on these implicit stability assertions.
I would rather just count how many sessions are returned with the same username and return the number of hits. If we have two sessions by the same user, `fortigate_vpn_users` would return 2 for that user. 

For `fortigate_vpn_users_receive_bytes_total` and similar traffic metrics we can't just add them all together (due to how metric counters work) - so there we need to have something more elaborate and per session (assuming the Fortigate returns data transferred per VPN connection).

I would probably split this up in two PRs, one for fixing `fortigate_vpn_users` - one for adding the data transferred.

IMSdevsecu:
Thanks for your feedback @bluecmd 
Ok for not removing metric until next major version. I understand. Will do a new cleanup PR for that specific point.
Will do a new PR to discuss how we can expose transmited/received VPN traffic.

bluecmd (approved):

bluecmd:
@secustor Please have a look

PR #133, opened 2021-08-27
feat(probe): Add probe for Fortimanager connection and registration status

IMSdevsecu:
Hi,

This code adds support for probing Fortimanager connection and registration status.
This metric helps us to detect issue with Fortimanager connection status.

I have not added `server` as it would be more helpful for compliance than for monitoring.
Do you think `server` information can be useful? Could be added later anyway.


secustor (dismissed):

pkg/probe/system_fortimanager_test.go
```diff
@@ -0,0 +1,31 @@
...
+	c := newFakeClient()
+	c.prepare("api/v2/monitor/system/fortimanager/status", "testdata/system-fortimanager-status.jsonnet")
+	r := prometheus.NewPedanticRegistry()
+	if !testProbe(probeSystemFortimanagerStatus, c, r) {
+		t.Errorf("probeSystemFortimanagerStatus() returned non-success")
+	}
+
+	em := `
+	# HELP fortigate_fortimanager_connection_status Fortimanager status ID
+        # TYPE fortigate_fortimanager_connection_status gauge
```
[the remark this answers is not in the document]
IMSdevsecu: I missed it at the review, sorry about that

pkg/probe/system_fortimanager.go
```diff
@@ -0,0 +1,46 @@
...
+		FortimanStat_id = prometheus.NewDesc(
+			"fortigate_fortimanager_connection_status",
+			"Fortimanager status ID",
+			[]string{"vdom", "mode"}, nil,
+		)
+		FortimanReg_id = prometheus.NewDesc(
+			"fortigate_fortimanager_registration_status",
+			"Fortimanager registration status ID",
+			[]string{"vdom", "mode"}, nil,
+		)
```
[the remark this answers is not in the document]
IMSdevsecu: Will do and adapt other parts of the code

pkg/probe/system_fortimanager.go
```diff
@@ -0,0 +1,46 @@
...
+	)
+
+	var res SystemFortimanagerStatus
+	if err := c.Get("api/v2/monitor/system/fortimanager/status", "vdom=root", &res); err != nil {
```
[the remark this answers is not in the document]
IMSdevsecu: When making my tests I actually switched to a firewall having only one root VDOM. I though I was testing a multi VDOM one. This is why I forced the query on the root VDOM. I though only root VDOM was returning fortimanager configuration.
Making the test again, I see that not only the root VDOM provide me feedback.
Will adapt the code and test accordingly.
@secustor Thank you for your feedback !

pkg/probe/system_fortimanager.go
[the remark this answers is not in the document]
IMSdevsecu: I am going to let `mode` label to allow filtering on `normal`/`backup` mode

pkg/probe/system_fortimanager.go
```diff
@@ -0,0 +1,43 @@
...
+	Results SystemFortimanagerResults `json:"results"`
+	VDOM    string                    `json:"vdom"`
+}
+
+func probeSystemFortimanagerStatus(c http.FortiHTTP, meta *TargetMetadata) ([]prometheus.Metric, bool) {
+	var (
+		FortimanStat_info = prometheus.NewDesc(
+			"fortigate_fortimanager_info",
+			"Fortimanager infos",
+			[]string{"vdom", "mode", "connection_status", "registration_status"}, nil,
```
bluecmd, line 27: I think it would be better to have the `_status` ones as their own metrics given that they are integers and I assume referencing an enum value?

README.md
```diff
@@ -29,6 +29,8 @@ Supported metrics right now as follows.
 
 Global:
 
+ * _System/Fortimanager/Status_
```
bluecmd, line 32: Isn't this per-VDOM?

pkg/probe/system_fortimanager.go
bluecmd: Ah I see that @secustor commented the opposite. The way I see it the info metrics are a hack to put in useful information that can't be represented as a time series (or makes no sense to do so). Status surely makes sense, no?

README.md
bluecmd: I see that it was global first, so this is probably a left-over.

bluecmd:
My apologies, I somehow missed the previous review when reviewing this so you got some conflicting comments. We will work it out, but yeah it became a bit messy.

README.md
IMSdevsecu: Yes I made a mistake querying a non multiVDOM firewall. So I though this was a global metric, but this actually a per-VDOM.
Will fix that with next correction

pkg/probe/system_fortimanager.go
bluecmd: I thought of another reason: when you put things in labels they should ideally never change due to how Prometheus treats metric staleness and how it is optimized to keep a single metric low-cost.

This all assumes that the status fields are Enums - the fact that they are called `status_id` makes me not 100% sure of this though.

If it is indeed an enum, it would take something like 10 minutes until the old value would be considered stale and removed from the query response.

IMSdevsecu: @bluecmd I made this on the first place, but I am not mastering the metric/label optimization so I relyed on @secustor experience.
I also think that info metric are a kind of hack for non numeric information.
Here connection_status and registration_status are 2 int, therefore I think a dedicated metric is better to graph evolution.
The only thing that could make a point against 2 dedicated metrics is the fact those 2 metrics will change unfrequently.
They are actually enum:
image

Just asking confirmation to save your time for next PR 😃 
I was wondering what is more cost effective between creating a new dedicated metric vs adding a label to an existing one ?
In our example, is it more storage efficient to store 2 metrics rather than 1 info metric with 2 labels that can change ?
My guess is if a label do not change for a very long time, then the impact is minimal on the metric as it is not stored for every scrape.

Also is there a rule or a limit about the number of labels to add to a metric ?
This refer to the issue of adding many labels to a metric (you discuraged me from adding to many labels on #127)

bluecmd: Kind of unrelated, but @IMSdevsecu that API documentation looks very nice - is that Fortinet Developer Network or something else?

Since you see that it is an enum and you can possibly browse the actual states, there is an even better way you can do:

```
fortigate_virtual_wan_status{interface="wan2",sla="Internet Check",state="disable",vdom="root"} 1
fortigate_virtual_wan_status{interface="wan2",sla="Internet Check",state="down",vdom="root"} 0
fortigate_virtual_wan_status{interface="wan2",sla="Internet Check",state="error",vdom="root"} 0
fortigate_virtual_wan_status{interface="wan2",sla="Internet Check",state="unknown",vdom="root"} 0
fortigate_virtual_wan_status{interface="wan2",sla="Internet Check",state="up",vdom="root"} 0
```

This pattern is pretty nice as it works around the issues about stale metrics while still allowing the use of a label with a human readable enum name. If we have those, which you might have from the view of the screenshot, that could be a really neat user experience.

As for metrics - the rule to think about is that only the value is the thing that can change in a time series / metric. If the label changes it becomes a totally different metric in the database and subject to all what that entails. Storage efficiency might be a concern for large metrics, but in practice the staleness issue is the way more annoying one.

Imagine if you were to write an alert for a Fortimanager being unavailable. Let's assume that is status=1. You would think you could do something like `fortigate_fortimanager_info{status="1"} == 1` right? That would work for alerting when that state happens, but you would **also** have the previous value (maybe `fortigate_fortimanager_info{status="2"} 1`) for a while. So if you for example did `count(fortigate_fortimanager_info)` you would receive 2 for the staleness duration. That's why the "enum expansion" pattern above works, it specifically is made with the idea that all enum values are always present, but only one of them is every 1 at a scrape. That way you can do things like `fortigate_fortimanager_info{status="down"} == 1` and it will work exactly like you want.

IMSdevsecu: @bluecmd Yes this is an extract from FNDN, I guess you do not have access as you are asking 😄
Let me know if you need confirmation from the API documentation later, I can have a look for you.
It is far from being perfect but it can help sometimes when this is documented enough.

Thank you the detailed explanation, this is cristal clear now 👌 

Unfortunatly, I just have the enum value in the API documentation, not the equivalent name.

IMSdevsecu: I have found no documentation giving the name, I can just tell that :
- status:
  - 2=Up
- registration:
  - 2=Registered

I will not be able to find all other values and I have no Fortimanager/fortigate to run test on.
What do you think is the best solution from there? 2 distinct metrics with id as value ? 1 info metric with id ?

bluecmd: That's a shame, oh well. 2 metrics with the id as value is the best in this case I'd say.

IMSdevsecu: Before we stay with this non optimal option, I am going to try to get the information from the Fortinet support.
Not sure this will help, but just trying this last attempt.

IMSdevsecu: Finally got a feeback from Fortinet support 🎉. I patched the code with the value meaning and splited it back into 2 metrics

bluecmd (approved):

PR #137, opened 2021-08-31
feat(probe): Add probe for fortigate internal clock time

IMSdevsecu:
Hi,

This metric is interesting for detecting internal clock badly synchronized.
This NTP synchro status is not present in the API of old FortiOS, so there is no other way to monitor clock drift in this case (at least none I am aware of 😄 )

I limited the request to the root VDOM, as time is shared accross all VDOM.
For the metric type I have chosen to stay with Gauge. Counter could habe been a better choice, but I though some use case can break the always growing value:
- clock is resynchronized backward, then value will decrease
- firewall is restarted and start back with the default hour (if cell kipping time is dead for example)
- firewall is replaced by a new one after a failure. The new one can not be synchronized yet

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ Global:
 
  * _System/Status_
    * `fortigate_version_info`
- * _System/Time_
+ * _System/Time/Clock_
    * `fortigate_time_seconds`
  * _System/Resource/Usage_
    * `fortigate_cpu_usage_ratio`
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ Global:
  * _System/Status_
    * `fortigate_version_info`
  * _System/Time_
-   * `fortigate_time`
+   * `fortigate_time_seconds`
  * _System/Resource/Usage_
    * `fortigate_cpu_usage_ratio`
    * `fortigate_memory_usage_ratio`
--- a/README.md
+++ b/README.md
@@ -312,6 +312,7 @@ To improve security, limit permissions to required ones only (least privilege pr
 |System/LinkMonitor           | sysgrp.cfg         |api/v2/monitor/system/link-monitor |
 |System/Resource/Usage        | sysgrp.cfg         |api/v2/monitor/system/resource/usage |
 |System/Status                | *any*              |api/v2/monitor/system/status |
+|System/Time/Clock            | sysgrp.cfg         |api/v2/monitor/system/time |
 |System/VDOMResources         | sysgrp.cfg         |api/v2/monitor/system/resource/usage |
 |User/Fsso                    | authgrp            |api/v2/monitor/user/fsso |
 |VPN/IPSec                    | vpngrp             |api/v2/monitor/vpn/ipsec |
--- a/README.md
+++ b/README.md
@@ -32,6 +32,8 @@ Global:
 
  * _System/Status_
    * `fortigate_version_info`
+ * _System/Time_
+   * `fortigate_time`
  * _System/Resource/Usage_
    * `fortigate_cpu_usage_ratio`
    * `fortigate_memory_usage_ratio`
--- a/pkg/probe/probe.go
+++ b/pkg/probe/probe.go
@@ -119,6 +119,7 @@ func (p *ProbeCollector) Probe(ctx context.Context, target string, hc *http.Clie
 		{"System/LinkMonitor", probeSystemLinkMonitor},
 		{"System/Resource/Usage", probeSystemResourceUsage},
 		{"System/Status", probeSystemStatus},
+		{"System/Time", probeSystemTime},
 		{"System/VDOMResources", probeSystemVDOMResources},
 		{"User/Fsso", probeUserFsso},
 		{"VPN/IPSec", probeVPNIPSec},
--- a/pkg/probe/probe.go
+++ b/pkg/probe/probe.go
@@ -119,7 +119,7 @@ func (p *ProbeCollector) Probe(ctx context.Context, target string, hc *http.Clie
 		{"System/LinkMonitor", probeSystemLinkMonitor},
 		{"System/Resource/Usage", probeSystemResourceUsage},
 		{"System/Status", probeSystemStatus},
-		{"System/Time", probeSystemTime},
+		{"System/Time/Clock", probeSystemTime},
 		{"System/VDOMResources", probeSystemVDOMResources},
 		{"User/Fsso", probeUserFsso},
 		{"VPN/IPSec", probeVPNIPSec},
--- a/pkg/probe/system_time.go
+++ b/pkg/probe/system_time.go
@@ -1,0 +1,38 @@
+package probe
+
+import (
+	"log"
+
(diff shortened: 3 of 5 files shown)


README.md
```diff
@@ -31,6 +31,8 @@ Global:
 
  * _System/Status_
    * `fortigate_version_info`
+ * _System/Time_
+   * `fortigate_time`
```
bluecmd, line 35:
```suggestion
   * `fortigate_time_seconds`
```

FWIW, the node_exporter has a metric named `node_time_seconds`.

README.md
```diff
@@ -31,6 +31,8 @@ Global:
 
  * _System/Status_
    * `fortigate_version_info`
+ * _System/Time_
```
bluecmd, line 34: I think we can keep this under Status to avoid having a bunch of categories with only one metric in them. We can move it to System/Status/Time later if we want to.

pkg/probe/system_time.go
```diff
@@ -0,0 +1,38 @@
...
+
+	"github.com/bluecmd/fortigate_exporter/pkg/http"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+func probeSystemTime(c http.FortiHTTP, meta *TargetMetadata) ([]prometheus.Metric, bool) {
+	var (
+		mTime = prometheus.NewDesc(
+			"fortigate_time",
+			"System epoq time in seconds",
```
bluecmd, line 14:
```suggestion
			"System epoch time in seconds",
```

bluecmd (changes requested):
Regarding `gauge` - that is fine, system wall-clock time can indeed move backwards under normal operations so it should be a gauge.

pkg/probe/system_time.go
```diff
@@ -0,0 +1,38 @@
...
+	"log"
+
+	"github.com/bluecmd/fortigate_exporter/pkg/http"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+func probeSystemTime(c http.FortiHTTP, meta *TargetMetadata) ([]prometheus.Metric, bool) {
+	var (
+		mTime = prometheus.NewDesc(
+			"fortigate_time",
```
IMSdevsecu, line 13:
```suggestion
			"fortigate_time_seconds",
```

pkg/probe/system_time_test.go
```diff
@@ -0,0 +1,28 @@
...
+	c.prepare("api/v2/monitor/system/time", "testdata/system-time.jsonnet")
+	r := prometheus.NewPedanticRegistry()
+	if !testProbe(probeSystemTime, c, r) {
+		t.Errorf("probeSystemTime() returned non-success")
+	}
+
+	em := `
+	# HELP fortigate_time System epoq time in seconds
+	# TYPE fortigate_time gauge
+	fortigate_time 1.630313596e+09
```
IMSdevsecu, line 22:
```suggestion
	# HELP fortigate_time System epoch time in seconds
	# TYPE fortigate_time gauge
	fortigate_time_seconds 1.630313596e+09
```

README.md
IMSdevsecu: For now the probe category name inherit the API URL path, making `System/Time`, `System\Status` or `System\Status\Time` break this naming convention.
Gathering mulitple metrics in a category name can be sometimes easier for inclusion/exclusion but remove the maximum granularity for inclusion/exclusion that we have right now.
Kipping a link between API URL path and category name maximize optimization when including/excluding probes as a API call is done or not depending on a probe being included/excluded.
Therefore no useless API call are done.

That being said, even if I believe in KISS, I am leaning for `System/Status/Time` which keep the link between API call and metric name. Even if this is not exactly the same path, at least there is a dedicated category name.
This has the advantage of permitting prefix matching.
Seems acceptable as we already have small arrangements for example with `BGP/Neighbors/IPv4` 😏
It does not remove a new category name tho. 😏

Do we go for `System/Status/Time` then ?

bluecmd: We can go with either `System/Time` or `System/Status/Time` - you make a good point with the API naming. I will leave the choice to you. We might want to expose things like NTP status later so we might add more metrics under the category either way.

IMSdevsecu: FYI, in FortiOS v7 NTP has URL path `system/ntp/status`

I think we should tweek probe category name following those rules:
- By default, probe category name should be derived from the API URL path (ex: `api/v2/monitor/user/fsso` become `User/Fsso`)
- A single probe category name should never be a subpart of another probe category name, in order to keep inclusion/exclusion declaration simple.
- probe category name should match : `^([a-ZA-Z0-9]+/)+[a-ZA-Z0-9]+$`
- Removing a prefix part is possible if not conflicting or bringing ambiguity with current and futur probe category name/URL path (ex: `BGP/Neighbo...` instead of `Router/BGP/Neighbo...` -- shorter is better
- Adding a suffix part to avoid prefix matching unwilled inclusion/exclusion is possible (ex: `api/v2/monitor/vpn/ssl` and `api/v2/monitor/vpn/ssl/stats` goes to `VPN/Ssl/Connections` and `VPN/Ssl/Stats`)
- Creating an additional category section not present in the API URL path is possible to make grouping more efficient and anticipate present or futur Fortigate API prefix match issues (ex: `monitor/system/time` conflict with `monitor/system/timezone`, then category name can be `System/Time/Clock` and `System/Time/Timezone`)

I agree that grouping in categories is a good thing. For now probe category name look well sized to me and do respect above rules.
About grouping all time related probes in a category like `System/Time/...` (ex: `System/Time/Clock` and `System/Time/NTP/Status`) actually looks like the best solution. `system/Timezone` made up my mind 😄 (even if `timezone` has very few interest, it shows the kind of problem that could come to us with only URL path rule). 
I think, on the long term, the above rules make things simplier even if we do not 100% follow the Fortigate API URL path we are still pretty close to.  

Will propose a draft PR for Contributing.md to cast those rules in stone 🎵 (#64)

Finally going for `System/Time/Clock`

WDYT @bluecmd ?

README.md
```diff
@@ -32,6 +32,8 @@ Global:
...
```
IMSdevsecu, line 35:
```suggestion
 * _System/Time/Clock_
```

pkg/probe/probe.go
```diff
@@ -119,6 +119,7 @@ func (p *ProbeCollector) Probe(ctx context.Context, target string, hc *http.Clie
 		{"System/LinkMonitor", probeSystemLinkMonitor},
 		{"System/Resource/Usage", probeSystemResourceUsage},
 		{"System/Status", probeSystemStatus},
+		{"System/Time", probeSystemTime},
```
IMSdevsecu, line 122:
```suggestion
		{"System/Time/Clock", probeSystemTime},
```

README.md
bluecmd: Sounds good to me :+1: @secustor any issues with the above?

secustor: @bluecmd  I have nothing to add to the discussion :+1:

bluecmd (approved):

PR #142, opened 2021-09-03
refactor(config): config generation and add unit testing

secustor:
- use flag.func instead of manual parsing
- add unit tests
- add snapshot tests
- replace global config with parameter 
- improve Makefile


bluecmd:
Can you break out the coverage and makefile stuff to its own PR? I think we can merge that pretty much straight away, the other things will need some time to look over.

internal/utils/test/snapshot.go
```diff
@@ -0,0 +1,10 @@
+package test
+
+import (
+	"path"
+
+	"github.com/bradleyjkemp/cupaloy/v2"
+)
+
+var Snapshotting = cupaloy.New(
+	cupaloy.SnapshotSubdirectory(path.Join("testdata", "snapshots")))
```
bluecmd, line 10: I think `path/filepath` is what we want to use here.

secustor: Yeah, after reading up the documentation this should be the more stable choice.    
I have replaced all usages of `path` regarding file operations.

.github/workflows/go.yml
```diff
@@ -25,7 +25,7 @@ jobs:
 
       - name: Ensure that all files are properly formatted
         run: |
-          FILES=$(go run golang.org/x/tools/cmd/goimports -w -l .)
+          FILES=$(make fmt-fix)
```
bluecmd, line 28: I think you had `run: make fmt-check` here before and it was very neat looking - did it not work as you wanted? Or did I make that up?

internal/config/main.go
```diff
@@ -52,82 +66,58 @@ type LocalCert struct {
 	Content []byte
```
bluecmd, line 66: Do you want to change this to `string` perhaps given that it's 1) base64 encoded so it doesn't matter, and 2) it is part of the snapshots so it looks a bit messy with the hexdump since it's only ASCII data?

internal/config/main_test.go
```diff
@@ -0,0 +1,159 @@
...
+)
+
+var (
+	validAuthFile      = test.GetFixturePathPanic("validAuthFile.yaml")
+	otherValidAuthFile = test.GetFixturePathPanic("otherValidAuthFile.yaml")
+	invalidAuthFile    = test.GetFixturePathPanic("invalidAuthFile.yaml")
+	certFile           = test.GetRelativeFixturePathPanic("cert.pem")
+)
+
+func TestInit(t *testing.T) {
```
bluecmd, line 21: This is 3 functions smashed into one it seems, can we split it to `TestInit_AuthFile` / `TestInit_CACerts` etc?

internal/utils/files/main.go
```diff
@@ -3,15 +3,15 @@ package files
 import (
```
bluecmd, line 3: I must have missed this file in a previous PR, I see it's not new - but could you add a header to why we have this file? I am confused about why it is needed. If it is only needed for the fixtures lookup, consider moving it to that file.

internal/utils/test/fixtures.go
```diff
@@ -0,0 +1,42 @@
...
+	"log"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/bluecmd/fortigate_exporter/internal/utils/files"
+)
+
+func GetFixturePath(fixtureFileName string) (string, error) {
+	callerDir, err := files.GetCallerDir(2)
```
bluecmd, line 13: Don't use magic numbers like "2". It seems like we use it a lot in this code here, so you probably want to define an alias or helper function.

.github/workflows/go.yml
secustor: I had it this way at some point. But this way we pretty print the offending files

internal/config/main.go
secustor: I would simply keep it that, as the snapshot is not to be read usually by humans

internal/config/main_test.go
secustor: Sadly not because this has to be executed serially, as I modify the global args and flag objects

internal/utils/files/main.go
secustor: It has been a part of `probe_test` to read out the jsonnet files, but after 5dcbf7324ee258c0fc0b99cad87cd517b6bc26a7 this has not be needed anymore. I have no special feelings regarding to keep it there or move it. If you prefer to move it all to `internal/utils/fixture` that's fine for me.

internal/utils/test/fixtures.go
secustor: `2` means in this context simply the function above me. 

```
aMethod --> GetFixturePath --> GetCallerDir
    2               1                0
```

What would you name the constant? I have no really nice constant name right now. 

The other alternative would a default value: 
```golang 
func GetRelativeFixturePathPanic(fixtureFileName string, additionalParameters ...interface{}) string {
	stackTraceStep := 2
	if len(additionalParameters) >= 1{
		stackTraceStep = additionalParameters[1].(int)
	}
	callerDir, err := files.GetCallerDir(stackTraceStep)
	if err != nil {
		log.Panicf("failed to get caller stack %v", err)
	}
	workDir, err := os.Getwd()
	if err != nil {
		log.Panicf("failed to get workdir %v", err)
	}
	relDir := strings.TrimPrefix(callerDir, workDir)
	fixturePath := filepath.Join(relDir, "testdata", "fixtures", fixtureFileName)
	return fixturePath
}

```

bluecmd: I would probably write it like this:

```go

type Depth int

const (
        ONE_UP Base = 2
        TWO_UP Base = 3
        ...
)

func MustFixturePath(depth Depth) string {
	_, filename, _, ok := runtime.Caller(int(depth) /* stack depth to descend */)
	if !ok {
                panic(...)
	}
	workDir, err := os.Getwd()
	if err != nil {
		log.Panicf("failed to get workdir %v", err)
	}
	relDir := strings.TrimPrefix(path.Dir(filename), workDir)
	return filepath.Join(relDir, "testdata", "fixtures")
}
```

Though - do we actually need to construct these? Can't we

internal/config/main_test.go
bluecmd: Can we figure out a way that breaks out of the mess of doing global state? It really would improve the current exporter code a lot.

internal/config/main.go
```diff
@@ -52,82 +66,58 @@ type LocalCert struct {
...
 }
 
-var (
-	parameter = FortiExporterParameter{
-		AuthFile:      flag.String("auth-file", "fortigate-key.yaml", "file containing the authentication map to use when connecting to a Fortigate device"),
+func setParameterFlags() *FortiExporterParameter {
```
bluecmd, line 69: From the other comment you said you modify global flag state. I assumed the goal of this code here was to remove depdendency on global flag state, if it's not - why move the flag initializers into this function instead of using the pretty common pattern of defining them in the file's `var` as we did before?

secustor: The target of this refactoring is to remove the global config state, which allows to drop this `reinit` logic altogether and ease the use of custom config values during testing.     

This is used for testing so that all flag pointers can be be recreated.     
What I referencing with global flag and args is the attributes of the imported packages `flag.CommandLine` and `os.Args`     

It hasn't been in scope for this PR to use Flagsets instead of flag, but I will look into it.

internal/utils/test/fixtures.go
secustor: As long you run `go test ./...` or you run test in the directory, we are fine removing then.

bluecmd:
Closing due to inactivity

PR #143, opened 2021-09-03
ci(Makefile): add coverage report and run additional tests

secustor:
Context:  https://github.com/bluecmd/fortigate_exporter/pull/142#issuecomment-912801357

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,10 +40,6 @@ test-unittests:
 test: fmt-check vet test-unittests generate-html-coverage print-coverage
 	@printf "Sucessfully run tests \n"
 
-.PHONY: test-update-snapshots
-test-update-snapshots:
-	UPDATE_SNAPSHOTS=true go test -v -race ./...
-
 .PHONY: get-dependencies
 get-dependencies:
 	go get -v -t -d ./...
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -23,21 +23,46 @@ build-release-arm64:
...
+test-unittests:
+	go test -v -race -coverprofile cover.out ./...
+
 .PHONY: test
-test:
-	go test -v -race ./...
+test: fmt-check vet test-unittests generate-html-coverage print-coverage
+	@printf "Sucessfully run tests \n"
+
+.PHONY: test-update-snapshots
```
bluecmd, line 43: This is for a later PR no?

Makefile
```diff
@@ -23,21 +23,46 @@ build-release-arm64:
...
 fmt-fix:
-	go run golang.org/x/tools/cmd/goimports -w -l .
+	@go mod download golang.org/x/tools
+	@go run golang.org/x/tools/cmd/goimports -w -l .
+
+.PHONY: fmt-check
+fmt-check:
+	@printf "Check formatting... \n"
+	@go mod download golang.org/x/tools
+	@if [[ $$( go run golang.org/x/tools/cmd/goimports -l . ) ]]; then printf "Files not properly formatted. Run 'make fmt-fix' \n"; exit 1; else printf "Check formatting finished \n"; fi
```
bluecmd, line 68:
```suggestion
	@go run golang.org/x/tools/cmd/goimports -l . || (printf "Files not properly formatted. Run 'make fmt-fix' \n"; exit 1) && printf "Check formatting finished \n"
```

I haven't tested that, but I'm pretty sure you don't need to use a whole if statement there.

secustor: goimports doesn't return a non 0 exit code if it finds formatting changes. Because of this doesn't work. Instead I'm looking for an output.  To be sure I have tried it, but the command returns in every case success.

Makefile
secustor: Yeah, that's correct. This is for the PR which reworks the config handling

bluecmd (approved):

PR #176, opened 2022-07-29
Removed Continue on DialUp VPNs

### LINKED ISSUE #170: No metrics collected for dialup IPsec VPNs
Hi and thank you very much for the great project.

Unfortunately I have noticed that all IPsec VPNs of type "dialup" are skipped entirely during metric collection:

https://github.com/bluecmd/fortigate_exporter/blob/81adda5136cc5f9160c0ec7c9f10ccdcd63713a4/pkg/probe/vpn_ipsec.go#L55-L61

This obviously means that any dialup VPN cannot be monitored, which is the majority in our environment.
I confirmed with a manual `curl` API call that all our VPNs are returned by the `/api/v2/monitor/vpn/ipsec` endpoint, so it's just this exporter that then filters them out.

Could this continue statement be removed so that metrics for all IPsec VPNs are collected? I can make a PR for it too if you prefer.

badbart:
Dialup VPNs are not only used with Client RemoteAccess VPNs, so removed the `continue`. 

It will fix #170

### DIFF
--- a/pkg/probe/vpn_ipsec.go
+++ b/pkg/probe/vpn_ipsec.go
@@ -51,14 +51,7 @@ func probeVPNIPSec(c http.FortiHTTP, meta *TargetMetadata) ([]prometheus.Metric,
 
 	m := []prometheus.Metric{}
 	for _, v := range res {
-		for _, i := range v.Results {
-			/*
-			  type 'dialup' seems to be client vpn.
-			  Not sure exactly what the difference is between probeVPNSsl
-			*/
-			if i.Type == "dialup" {
-				continue
-			}
+		for _, i := range v.Results {			
 			for _, t := range i.ProxyID {
 				s := 0.0
 				if t.Status == "up" {


bluecmd:
Hi, thanks for the contribution!

I am mostly concerned that removing this guard clause will cause metric duplication. Can you contribute a JSON output from the API with a configuration that has both the existing supported VPN (SSL I suppose?) and the one you want? That way we can ensure that we do not regress the existing support.

badbart:
Hey, just send the JSON to you via Mail.

cloudistration-admin:
Hello, when will this be fixed?

bluecmd:
I need somebody that has both VPN types to convince me that this doesn't break the existing users. I haven't been able to take time to do this myself.

jantari:
I opened the original issue (#170) and I have a 600E with regular IPSec, Dialup IPsec and SSLVPN all configured simultaneously.

What can I do to help?

PR #184, opened 2022-08-07
feat(probe): Add system reboot and snapshot time

### LINKED ISSUE #174: Fortigate system uptime as a metric
Hello,

Perfect exporter and works very well, thanks.
Would it be possible to get the fortigates system uptime?

Cheers,
Heini

bluecmd:
Fixes #174 

I couldn't find any other source of this information than in the web-UI state document. It seems to be stable from 6.0 and to 7.0 at least so we should be able to use it - even though the internal naming becomes a bit awkward. This document also seems to be undocumented from what I can tell, so I expect this to be a last-resort / best-effort kind of place to grab metrics from.


pkg/probe/webui_state.go
```diff
@@ -0,0 +1,45 @@
...
+	type WebUIState struct {
+		SnapshotUTCTime float64 `json:"snapshot_utc_time"`
+		UTCLastReboot   float64 `json:"utc_last_reboot"`
+	}
+
+	type webuiState struct {
+		Results WebUIState
+	}
+
+	var sstate webuiState
```
secustor, line 33:
```suggestion
	var state webuiState
```
Maybe use `state` instead of `sstate`, at least for me it is not clear what the first s stands for. ;)

pkg/probe/webui_state.go
```diff
@@ -0,0 +1,45 @@
...
+
+	if err := c.Get("api/v2/monitor/web-ui/state", "", &sstate); err != nil {
```
secustor, line 35:
```suggestion
	if err := c.Get("api/v2/monitor/web-ui/state", "", &state); err != nil {
```

pkg/probe/webui_state.go
```diff
@@ -0,0 +1,45 @@
...
+		log.Printf("Error: %v", err)
+		return nil, false
+	}
+
+	m := []prometheus.Metric{
+		prometheus.MustNewConstMetric(mRebootTime, prometheus.GaugeValue, sstate.Results.UTCLastReboot/1000),
+		prometheus.MustNewConstMetric(mSnapshotTime, prometheus.GaugeValue, sstate.Results.SnapshotUTCTime/1000),
```
secustor, line 42:
```suggestion
		prometheus.MustNewConstMetric(mRebootTime, prometheus.GaugeValue, state.Results.UTCLastReboot/1000),
		prometheus.MustNewConstMetric(mSnapshotTime, prometheus.GaugeValue, state.Results.SnapshotUTCTime/1000),
```

secustor (approved):
A small nitpick else LGTM

PR #206, opened 2023-03-10
Add support for "dynamic" configuration where token and "profile" can…

thenodon:
… be query parameters, where profile is an entry in fortigate-key.yaml to get the probes exclude/include definition.

Add support for Fortigate managed switches metrics called Switch/ManagedSwitch. The metrics are collected from endpoint api/v2/monitor/switch-controller/managed-switch.


thenodon:
@bluecmd would you consider this PR?

bluecmd:
@thenodon Thanks for the PR - looks good from what I can see.

Could you:
1. Open a PR for the managed switched stuff separately? I prefer to keep features and refactors separate in the git log if possible. I did a quick scan of the metrics and they look good, but I will need to take a more careful look at those given that they are a pain to change once they are released and people start using them.
2. I am personally not a fan of passing the token as a URL parameter, but I will not tell people how to run their infra. I would however ask you  to add a notice about that passing the token over `http` as opposed to `https` is insecure, and the fact that the token is presented in clear text in Prometheus, both in the config as well as the Web UI under scraping targets, would have to be considered. If you add such a notice I don't see any issues with merging this part.

Regards,

thenodon:
Thanks for the response @bluecmd. I will separate the PR in two as you suggested. Regarding bullet 2 I have added a "warning" text in the README.md in the PR like:

Make sure to use the last labeldrop on the `token` label so that the tokens is not be part of your time series.
>Since `token` is a label it will be shown in the Prometheus webgui at `http://<your prometheus>:9090/targets`.

> **Make sure you protect your Prometheus if you add the token part of your prometheus config** 
> 
> Some options to protect Prometheus:
> - Only expose UI to localhost --web.listen-address="127.0.0.1:9090"
> - Basic authentication access - https://prometheus.io/docs/guides/basic-auth/
> - It is your responsibility!

Should also add use of https. And I understand your concern, but this is purely for managing automation related to service discovery. If we do not have the option to pass token from Prometheus the automation need also to include an update of the fortigate-exporter config and a restart. 
For discovery I have created https://github.com/thenodon/fortigate-exporter-discovery. Currently works for file service discovery, but will be possible to run in "server mode" providing http based service discovery.

bluecmd:
@thenodon Is there anything else to be done in this PR? I see the two other PRs are merged. If nothing remains, please close this one.

thenodon:
@bluecmd - you have a busy day :) Sorry that I had forgot to close it

Issue #195, opened 2022-10-20
General improvements & Feedback


33Fraise33:
Hello,

I am implementing fortigate_exporter for one of our clusters. For us to replace our generic snmp scraping with this one completely there are some items that I bump into which might be fixable: 
- Allow to define a default set of include and exclude rules for all hosts
- Allow to define one api key for multiple hosts (in case we define the same api key through fortimanager for example) 
- For metrics like "fortigate_fortimanager_connection_status" it make more sense to have one metric with a value defining the status than having all the statuses but with a "true/false" value this also for storage usage
- Is it possible to add the interface admin status to fortigate_interface_link_up (to make an alert for: fortigate_interface_link_up{admin_status="true"}==0
- Is it possible to return bgp advertised routes to a neigbor

I will try to create a pull request where possible but I do not really have any go knowledge.

PR #214, opened 2023-04-18
Implemented feature request from #195

33Fraise33:
Hello,

I have requested in #195 some features that would improve the usability of Fortigate-exporter. Even though I do not know any GO I tried to implement these features where possible. Feedback welcome

- ` fortigate_bgp_neighbor_ipv4_info` now returns a value from 1-6 indicating the status of the bgp session. This allows for easier alerting (`fortigate_bgp_neighbor_ipv4_info{admin_status=true} !=6`) and is more in line with other exporters. 
- Added a description to `fortigate_ipsec_tunnel_up` indicating the meaning of 1 and 0
- Added a description to `fortigate_interface_link_up` indicating that this status does not take into account admin status
- Changed the way fortimanager status is returned, now the statusses are not hardcoded as labels but they are returned as values (like the bgp status). The value meaning have been added to the help description of the metric. 
- Changed the way targets are formatted to have better compatibility with other prometheus jobs without needing specific rewrites. A target is now declared without "https", this is being added in code during probe
- Added the option to specify `"*"` as a target in the yaml file as a catchall/fallback target. This for massive deployments. There is room for improvement on this implementation, I would prefer to parse the yaml key as a regex giving the room for items like: *.domain.tld etc but my Go understanding was coming short here.

### DIFF
--- a/pkg/http/main.go
+++ b/pkg/http/main.go
@@ -19,9 +19,9 @@ type FortiHTTP interface {
 
 func NewFortiClient(ctx context.Context, tgt url.URL, hc *http.Client, aConfig config.FortiExporterConfig) (FortiHTTP, error) {
 
-	auth, ok := aConfig.AuthKeys[config.Target(tgt.String())]
+	auth, ok := aConfig.AuthKeys[config.Target(tgt.Host)]
 	if !ok {
-		return nil, fmt.Errorf("no API authentication registered for %q", tgt.String())
+		return nil, fmt.Errorf("no API authentication registered for %q", tgt.Host)
 	}
 
 	if auth.Token != "" {
@@ -34,7 +34,7 @@ func NewFortiClient(ctx context.Context, tgt url.URL, hc *http.Client, aConfig c
 		}
 		return c, nil
 	}
-	return nil, fmt.Errorf("invalid authentication data for %q", tgt.String())
+	return nil, fmt.Errorf("invalid authentication data for %q", tgt.Host)
 }
 
 func Configure(config config.FortiExporterConfig) error {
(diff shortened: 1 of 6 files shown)


bluecmd:
Hi!

The description changes looks fine, happy to merge those.

> Changed the way targets are formatted to have better compatibility with other prometheus jobs without needing specific rewrites. A target is now declared without "https", this is being added in code during probe

Sadly I think that's a breaking change, so we would have to support both ways - i.e. only add https:// unless there is a scheme present already.

> Added the option to specify "*" as a target in the yaml file as a catchall/fallback target. This for massive deployments. There is room for improvement on this implementation, I would prefer to parse the yaml key as a regex giving the room for items like: *.domain.tld etc but my Go understanding was coming short here.

My main issue with this is that it is insecure by design to have a fallback. Example: You add a fallback on your service and me as an attacker comes along and calls `/probe?target=my-ip-here` and then you will happily send me the token that is stored. This is why we are currently using opt-in style configuration.

> Changed the way fortimanager status is returned, now the statusses are not hardcoded as labels but they are returned as values (like the bgp status). The value meaning have been added to the help description of the metric.

These are also breaking changes. I am open for adding an additional metric variation that behaves like this though.

33Fraise33:
Thanks for your swift comments! 

For the breaking changes having them as a separate metric is also an option. It is just a better practice for the Prometheus cardinality to return values instead of labels. 

For the catchall option, I understand your concern. I do hope most of the people run the exporter in their private network and not exposed. An option is to allow the regex/catchall through an extra startup option? -insecure-targets or similar.

33Fraise33:
Hello,

Was is the plan with this? I feel ok to run this branch in production but I would prefer to see this implemented merged into the main product? Currently missing bgp session alerting is a really important one so I would like to know if this is going to be added or not.

bluecmd:
Would you be fine with keeping the state label but /also/ add the value as the gauge? I'd be ok with that in terms of breakage. That, or create a new metric.

Changing the current one by dropping labels is out of the question, it will break anyone depending on it.

Next time it's better if you create six PRs with one commit in each, that makes it easy for me to merge the parts I'm fine with straight away. Right now this metric naming issue is holding the other changes hostage

PR #215, opened 2023-04-26
feat: Add support for managed switches

thenodon:


thenodon:
This PR replace the managed switch part of PR #206

bluecmd (approved):
Looks good to me, do you want the metrics to be named `..._bcast_` or do you prefer the more readable `..._broadcast_` et.al.?

thenodon:
I have tried to stay as close to the attribute naming used in the fortigate api response and I think the short naming is used. I'm fine with both bcast and broadcast. I think you need to make the decision what is most inline with naming convention used in the the rest of the exporter.

thenodon:
@bluecmd I'm not able to find any error running go fmt. Can you share what you see is failing and the check you run.

thenodon:
Sorry was on the wrong go version. Hope it works now.

bluecmd:
Thanks!

thenodon:
Thank you - now time for the next PR :)

PR #216, opened 2023-04-27
Add support for "dynamic" configuration

thenodon:
This is a replace of PR #206 including only the additional support for request url and generic configuration.
This PR do not break existing configuration and can be used in combination with existing configuration.
The change add the following optional query parameters:
```
token=<token>&profile=FGT40F
```
In the `fortigate-key.yaml` the above profile is define as a "normal" entry like:
```
FGT40F:
  probes:
    exclude:
      - Switch/ManagedSwitch
      - Wifi
```
With this configuration, an entry in  `fortigate-key.yaml` becomes a sort of "profile" that during the Prometheus discovery can be defined based of model, adom etc and what ever is available during the Prometheus discovery phase.

The existing way to configure and the above can of course coexists in the `fortigate-key.yaml` like:
```
FGT40F:
  probes:
    exclude:
      - Switch/ManagedSwitch
      - Wifi

"https://localhost:443":
  token: XXXXXXXXXXXXXXXXXXXXXXXXXXXX
  probes:
    include:
      - System
```


thenodon:
@bluecmd to you feel it is okay? Have tried to do it as you requested in #206

bluecmd (approved):

bluecmd:
Looks good, thanks! :)

PR #219, opened 2023-05-10
Fortiswitch port stats

Jurka007:
@bluecmd or @secustor


bluecmd:
@Jurka007 can you rebase and fix the check errors?

README.md
```diff
@@ -51,7 +51,37 @@ Global:
...
+   * `fortiswitch_port_receive_packets_total`
+   * `fortiswitch_port_receive_bytes_total`
+   * `fortiswitch_port_receive_unicast_packets_total`
+   * `fortiswitch_port_receive_multicast_packets_total`
+   * `fortiswitch_port_receive_broadcast_packets_total`
+   * `fortiswitch_port_receive_errors_total`
+   * `fortiswitch_port_receive_drops_total`
+   * `fortiswitch_port_receive_oversized_packets_total`   
+ * _SwitchController/ManageSwitch/Health_
+   * `fortiswitch_health_summary_cpu`
```
bluecmd, line 75: * Avoid using words like mem when memory works
 * These health metrics names have no unit in them (e.g. add _celsius to`fortiswitch_health_temperature`

pkg/probe/fortiswitch_health.go
```diff
@@ -0,0 +1,151 @@
...
+		Summary           Summary           `json:"summary"`
+		Poe               Poe               `json:"poe"`
+	}
+
+	type swResponse struct {
+		Results map[string]Results `json:"results"`
+		VDOM    string
+	}
+	var r swResponse
+	//var r map[string]swResponse
```
bluecmd, line 119: Remove dead code

pkg/probe/fortiswitch_health.go
```diff
@@ -0,0 +1,151 @@
...
+	"github.com/bluecmd/fortigate_exporter/pkg/http"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+func probeSwitchHealth(c http.FortiHTTP, meta *TargetMetadata) ([]prometheus.Metric, bool) {
+	var (
+		mSumCPU = prometheus.NewDesc(
+			"fortiswitch_health_summary_cpu",
+			"Summary CPU health",
+			[]string{"rating", "fortiswitch", "VDOM"}, nil,
```
bluecmd, line 15: Lowercase all labels

pkg/probe/fortiswitch_health.go
```diff
@@ -0,0 +1,151 @@
...
+
+	if err := c.Get("api/v2/monitor/switch-controller/managed-switch/health", "vdom=root", &r); err != nil {
+		log.Printf("Error: %v", err)
+		return nil, false
+	}
+	m := []prometheus.Metric{}
+	//for _, sw := range r {
+	for fswitch, hr := range r.Results {
+
+		m = append(m, prometheus.MustNewConstMetric(mSumCPU, prometheus.GaugeValue, hr.Summary.CPU.Value, hr.Summary.CPU.Rating, fswitch, r.VDOM))
```
bluecmd, line 130: The rating likely shouldn't be a level ä label but rather a metric like "xxx_is_good". That, or an enum label on a `_rating` metric. Otherwise it's really hard to use it in alerts.

pkg/probe/fortiswitch_health.go
```diff
@@ -0,0 +1,151 @@
...
+			[]string{"rating", "fortiswitch", "VDOM"}, nil,
+		)
+		mSumMem = prometheus.NewDesc(
+			"fortiswitch_health_summary_mem",
+			"Summary MEM health",
+			[]string{"rating", "fortiswitch", "VDOM"}, nil,
+		)
+		mSumUpTime = prometheus.NewDesc(
+			"fortiswitch_health_summary_uptime",
+			"Summary Uptime",
```
bluecmd, line 24: If you have the documentation it would be really great if these descriptions could be made more verbose so I for example have any idea what these means. But if you don't, it's fine - sometimes we simply don't know

pkg/probe/fortiswitch_port_stats.go
```diff
@@ -0,0 +1,198 @@
...
+		m = append(m, prometheus.MustNewConstMetric(mSwitchStatus, prometheus.GaugeValue, swStatus, swr.VDOM, swr.Name, swr.FgPeerIntfName, swr.Connection, swr.State))
+
+		for _, pi := range swr.Ports {
+			pStatus := 0.0
+			if pi.Status == "up" {
+				pStatus = 1.0
+			}
+			m = append(m, prometheus.MustNewConstMetric(mPortStatus, prometheus.GaugeValue, pStatus, swr.VDOM, swr.Name, pi.Interface, pi.Vlan, pi.Duplex))
+			m = append(m, prometheus.MustNewConstMetric(mPortSpeed, prometheus.GaugeValue, pi.Speed*1000*1000, swr.VDOM, swr.Name, pi.Interface, pi.Vlan, pi.Duplex))
+
```
bluecmd, line 174: Having duplex as a label seems a bit off to me. Why not have it as a separate metric?

pkg/probe/fortiswitch_port_stats.go
```diff
@@ -0,0 +1,198 @@
...
+		return nil, false
+	}
+	m := []prometheus.Metric{}
+	//	for _, v := range r {
+	for _, swr := range r.Results {
+		swStatus := 0.0
+		if swr.Status == "Connected" {
+			swStatus = 1.0
+		}
+		m = append(m, prometheus.MustNewConstMetric(mSwitchStatus, prometheus.GaugeValue, swStatus, swr.VDOM, swr.Name, swr.FgPeerIntfName, swr.Connection, swr.State))
```
bluecmd, line 165: The state label here should likely be an enum label if the possible values of state is limited and known. It's hard to use it for anything useful otherwise.

pkg/probe/fortiswitch_port_stats.go
```diff
@@ -0,0 +1,198 @@
...
+	var (
+		mSwitchStatus = prometheus.NewDesc(
+			"fortiswitch_status",
+			"Whether the switch is connected or not",
+			[]string{"vdom", "name", "fgt_peer_intf_name", "connection_from", "state"}, nil,
+		)
+		mPortStatus = prometheus.NewDesc(
+			"fortiswitch_port_status",
+			"Whether the switch port is up or not",
+			[]string{"vdom", "name", "interface", "vlan", "duplex"}, nil,
```
bluecmd, line 20: Can a port only have one VLAN on a
Fortiswitch? That seems odd. What happens when the port is a VLAN trunk?

pkg/probe/fortiswitch_port_stats.go
```diff
@@ -0,0 +1,198 @@
...
+	"github.com/bluecmd/fortigate_exporter/pkg/http"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+func probeSwitchPortStats(c http.FortiHTTP, meta *TargetMetadata) ([]prometheus.Metric, bool) {
+	var (
+		mSwitchStatus = prometheus.NewDesc(
+			"fortiswitch_status",
+			"Whether the switch is connected or not",
+			[]string{"vdom", "name", "fgt_peer_intf_name", "connection_from", "state"}, nil,
```
bluecmd, line 15: This meric is not like the rest. The fgt_peer_intf_name and connection_from is not clear to me what this metric is. Care to add it to the description or as a comment in code?

bluecmd (changes requested):
Some metric naming and label comments

pkg/probe/fortiswitch_health.go
Jurka007: Unfortunately there is nothing in the docs :( I have no idea either, we can only guess..

Jurka007:
I am back baby. I don't remember what I have done here :) Will try to update asap so we can merge this.

pkg/probe/fortiswitch_health.go
Jurka007: Seems to be a summary of the overall status, good or not good 🤷

pkg/probe/fortiswitch_health.go
Jurka007: Right, we don't care about the numbers as there is no documentations what those really means. The fortios has own logic to either return good or not good, but no idea what are the other ratings.
Perhaps the alert should be based on any other than good value.

Jurka007: wait, prometheus has only numeric metrics.. 🤔

pkg/probe/fortiswitch_port_stats.go
Jurka007: State: `Authorization state of the FortiSwitch`

Discovered, DeAuthorized, Authorized

Jurka007:
Btw the https://github.com/prometheus-community/fortigate_exporter/pull/215 has the same api call. didn't notice until now 
so it's kinda duplicate 🤦

Jurka007:
Ok, this should be it. Added support for new api changes. Didn't test yet, don't have any firewall with os 7.6

Also maybe worth to merge the managed_switch probe and this one.

n3s0:
Is there anything I can do to test or change to get this functional?

PR #221, opened 2023-05-23
doc(readme): Elaborate which parts are user specified

faizan-syed:
Update readme.md with example how to use prometheues configuration as per user local enviromnent

### DIFF
--- a/README.md
+++ b/README.md
@@ -470,6 +470,7 @@ An example configuration for Prometheus looks something like this:
       - target_label: __address__
         replacement: '[::1]:9710'
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -470,6 +470,7 @@ An example configuration for Prometheus looks something like this:
       - target_label: __address__
         replacement: '[::1]:9710'
 ```
+In above configuration only the targets and the replacement values needs to be changed as per your environment. Where target is URL of the fortigate firewall and the replacement (at the buttom) will be the FQDN of system where the node exporter is running, example ```replacement: 'YourSystem.public.corp.com:9710'```
```
bluecmd, line 473:
```suggestion
In above configuration only the targets and the replacement values needs to be changed as per your environment. Where target is URL of the Fortigate firewall and the replacement (at the bottom) will be the FQDN of system where the node exporter is running, example ```replacement: 'YourSystem.public.corp.com:9710'```
```

bluecmd (approved):

PR #238, opened 2023-07-20
Changed return of fortimanager status to value

33Fraise33:
This change puts fortimanager status in line with bgp neighbour info, ipsec tunnel status,...


bluecmd:
As per https://github.com/bluecmd/fortigate_exporter/blob/master/CONTRIBUTING.md#naming-convention we should never break the API guarantee between the users by removing a label except if we ever do a 2.0 version. Sorry.

33Fraise33:
Hi @bluecmd how can we go ahead and set the target branch to a breaking change / future branch? As I feel that lowering the label cardinality is in favor of all users. 

The way the fortimanager one is written is not how you would normally return a value and causes annoyances during dashboarding and alertmanager queries. 
There is not one query returning the current status (which can be shown on a graph: `fortigate_fortimanager_registration_status`)

bluecmd:
I disagree that this type of cardinality increase is a problem (you're replacing one timeseries with X amounts (X being bounded and small), but each have 1/X activity so encoding wise they are much cheaper than the original one by themselves) - but it has been ages since I checked in with the Prometheus devs on this so I cannot claim to be knowledgeable on Prometheus internals anymore. If you feel strongly about this, I would encourage you to raise the topic on the Prometheus mailing list to solicit firm evidence that this way it is done today is not a good way to do metrics.

W.r.t. to not one query to return the status. Why wouldn't `sum(fortigate_fortimanager_registration_status) by (instance, status) != 0` work for you? That's generally how I display current enum status on e.g. Grafana.

33Fraise33:
Hi, except for the following I am currently not able to provide more documentation on this: https://prometheus.io/docs/practices/instrumentation/#do-not-overuse-labels & https://prometheus.io/docs/instrumenting/writing_exporters/#labels

It is not about being overused but just more optimised.

PR #259, opened 2023-12-01
feat: added ARP Table informatio and corresponding testcase

FriendlyButFire:
Files that have been added:
./pkg/probe/system_arp.go
./pkg/probe/system_arp_test.go
./pkg/probe/testdata/arp.jsonnet

Files that have been changed:
./pkg/probe/probe.go
./README.md

Tested on FortiGate KVM with FortiOS Version v7.0.5

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,9 @@ Per-VDOM:
  * _System/Fortimanager/Status_
    * `fortigate_fortimanager_connection_status`
    * `fortigate_fortimanager_registration_status`
+   * `fortigate_ippool_total_items`
+ * _System/Network/Arp
+   * `fortigate_arp_entry_age_seconds`
  * _System/Interface_
    * `fortigate_interface_link_up`
    * `fortigate_interface_speed_bps`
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -73,6 +73,9 @@ Per-VDOM:
  * _System/Fortimanager/Status_
    * `fortigate_fortimanager_connection_status`
    * `fortigate_fortimanager_registration_status`
+   * `fortigate_ippool_total_items`
```
bluecmd, line 76: Copy paste error?

bluecmd (commented):
This metric will have to be disabled by default if you really really want it in the current state. The metric dimensions are extremely large (any IP * any MAC * any interface).

Are you sure this is what you want? The total amount of MACs per interface isn't enough?

README.md
FriendlyButFire: yea, sorry, don't really know if to remove I have to recommit

FriendlyButFire:
> This metric will have to be disabled by default if you really really want it in the current state. The metric dimensions are extremely large (any IP * any MAC * any interface).
> 
> Are you sure this is what you want? The total amount of MACs per interface isn't enough?

Being disabled by default is fine for me, that's the use case I'm aiming for, I just wanted to share the implementation of it to everyone

PR #262, opened 2023-12-14
feat: added metrics for system vdom resources

xerox81:
Added vdom global resources metrics.

@bluecmd


pkg/probe/system_vdom_resources.go
```diff
@@ -0,0 +1,120 @@
...
+		}
+		o.Interfaces[k] = p
+	}
+	return nil
+}
+
+func probeSystemVDOMResourcesAll(c http.FortiHTTP, meta *TargetMetadata) ([]prometheus.Metric, bool) {
+	var (
+		vdomResources = prometheus.NewDesc(
+			"fortigate_vdom_resources",
```
bluecmd, line 81: These metrics need to be split up, one per metric. Having aggregated/"catch-all" ones like this makes little sense. See https://prometheus.io/docs/practices/naming/#metric-names, especially the last part: "...should represent the same logical thing-being-measured across all label dimensions. ".

Also:

> As a rule of thumb, either the sum() or the avg() over all dimensions of a given metric should be meaningful (though not necessarily useful). 

Doing `sum(fortigate_vdom_resources)` seems to me to not produce anything that makes any sense. Sticking values into the labels will explode cardinality as well as make it non-ideal to access the value for graphing purposes.

Finally, I cannot tell what the metrics are actually measuring from the name. E.g. what is `fortigate_vdom_resources{usage_percent}` or `fortigate_vdom_resources{custom_max}`?

bluecmd (changes requested):
Hi, thanks for your submission. I've added a comment around the metric structure.

pkg/probe/system_vdom_resources.go
lazyb0nes: > These metrics need to be split up, one per metric. Having aggregated/"catch-all" ones like this makes little sense. See https://prometheus.io/docs/practices/naming/#metric-names, especially the last part: "...should represent the same logical thing-being-measured across all label dimensions. ".
> 
> Also:
> 
> > As a rule of thumb, either the sum() or the avg() over all dimensions of a given metric should be meaningful (though not necessarily useful).
> 
> Doing `sum(fortigate_vdom_resources)` seems to me to not produce anything that makes any sense. Sticking values into the labels will explode cardinality as well as make it non-ideal to access the value for graphing purposes.
> 
> Finally, I cannot tell what the metrics are actually measuring from the name. E.g. what is `fortigate_vdom_resources{usage_percent}` or `fortigate_vdom_resources{custom_max}`?

I haven't yet verified what `usagepercent` is derieved from. I will have a look and get back on it but the practical idea here is if you let's say you have a big 3500 cluster for example and you have a maximum of let's say 35 million concurrent sessions. You'd ideally want to allocate sessions to each VDOM. 

Take in this example you'd allocate it to the Guaranteed number of sesisons, if you then want to make a Prometheus alert on it. You'd just divide the current usage by the Guaranteed to get the usage percentage. As it stands currently without this value you have to make a custom alert rule for each defined limit.

xerox81:
I have reduced the metric to only send session usage in percent per vdom because that is the only value I really needed. The metrics structure is better aligned with best practices.

lazyb0nes:
@bluecmd

adjosan:
What's missing?

PR #264, opened 2023-12-28
fix(probe): correct OSPF state enum

benclerc:
Hi @bluecmd !

Spotted a small issue in the OSPF code state returned, it appears to be shifted by 1 for each value.
Let me know if somethings not right. I assumed that the wanted value for the test was 8 (Full).

Benjamin


bluecmd (changes requested):
Can you add test cases for all states? That way we can see that this doesn't happen again. I see for example that Down returns 0, but the description says Down = 1. So might be good to just have tests for all states.

benclerc:
Good idea, just did !

bluecmd (approved):

benclerc:
Thank you for quick responses !

PR #267, opened 2024-01-04
feat: Added pba_per_ip return for ippool to differentiate between ippool types

33Fraise33:
Hello,

a while ago I added the ippool probe to the exporter. While using this we noticed that it is hard to see if an ip pool is exhausted if you are mix and matching one-to-one and pba ip pools. 
Unfortunately the api does not return the ippool type but it does return the amount of pba_per_ip which is 1 for a one-to-one and x for port-block-allocation. This way we can kind of differentiate the types being in use and we can create reasonable alerts: 
`fortigate_ippool_available_ratio < 0.2 and fortigate_ippool_pba_per_ip != 1`

Also it seems like overload ippools are not being returned by the API (which I have just noticed now).

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,7 @@ Per-VDOM:
    * `fortigate_ippool_clients`
    * `fortigate_ippool_used_items`
    * `fortigate_ippool_total_items`
+   * `fortigate_ippool_pba_per_ip`
  * _System/Fortimanager/Status_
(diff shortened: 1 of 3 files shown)


bluecmd:
Can you add it to the README as well? Thanks!

bluecmd (approved):

33Fraise33:
> Can you add it to the README as well? Thanks!

done!

PR #282, opened 2024-02-27
Update managed_switch.go

kristenungdom:
in order to get switch data after v. 7.2 you need to edit “managed_switch.go” to look like this: “api/v2/monitor/switch-controller/managed-switch/status”


maos999:
Hi, would appreciate if this PR could be approved since I've the same issue with v. 7.2 and this will be the fix

maos999 (approved):

antnsn (approved):
Looks good

antnsn (commented):
will work for new firmware, unsure how this work with earlier firmware?

maos999 (approved):
fix: looks good

PR #290, opened 2024-05-29
fix: update schema and path for managed switch

hexionas:


hexionas:
Actually, we have just found out that a lot of the expected port related metrics are missing. After some further investigation it seems that the `PortStats` is now derived from hitting the `api/v2/monitor/switch-controller/managed-switch/port-stats` endpoint as well as the `api/v2/monitor/switch-controller/managed-switch/status` endpoint and combining the results. I will take a look when I get a chance to get this to work and make the necessary changes.

PR #307, opened 2025-01-14
Update managed switch endpoint

antnsn:
Changed managed switch endpoint to work with fw 7.2
Added ghcr docker image

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -3,17 +3,48 @@ name: Docker
 on:
   push:
     branches: [ master ]
+    tags: [ 'v*.*.*' ]
   pull_request:
     branches: [ master ]
 
+env:
+  REGISTRY: ghcr.io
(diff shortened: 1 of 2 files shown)


yarons:
Hey, I'm not a project maintainer and I like and need your solution, let's start with prepending the term "fix: " to the beginning of the title.
So: `Update managed switch endpoint` -> `fix: Update managed switch endpoint`

This is required by Semantic PR, next, I guess you'd like to rebase your changes on top of the latest changes in main.

BTW is there a way to have a conditional endpoints according to the Fortigate firmware versions?

Let's hope this will be merged soon, int the meanwhile I found another malfunctioning endpoint and I can't find the full API implementation to see how to solve this, the endpoint is "api/v2/monitor/system/sensor-info" and I've already found which function is calling it but I'm not sure what for and what is the new endpoint replacing the malfunctioning one.

Thanks.

PR #324, opened 2025-06-30
fix(backend): update managed-switch API ...

antnsn:
fix(probe): support FortiGate firmware 7.2+ API changes

Update managed switch endpoint to api/v2 for firmware 7.2+ compatibility.


bastischubert:
with 6.4 (last version that will break with this change) going end of everything 2026-03-31 i think we should just merge it and risk braking ancient devices that are no longer supported, in favour of current systems.

thx for your contribution @antnsn

bastischubert (approved):

PR #334, opened 2025-09-10
Feature/dns latency

Philldomd:
New probe for dns latency, tested on fortiOS 7.4


SuperQ:
Almost there, just need to update the tests with `probeNetworkDNSLatency`.

PR #335, opened 2025-09-15
Feature/system ntp status

Philldomd:
Added probe for ntp status


SuperQ (approved):
Minor nit, otherwise LGTM.

PR #336, opened 2025-09-15
Feature/system performance status

Philldomd:
Adding performance status probe for all vdoms


Philldomd:
This one is broken. After testing this on 7.4 it seems that I interpreted the documentation wrong. Will commit an updated version shortly

PR #341, opened 2025-10-21
Update build

SuperQ:
* Update minimum Go to 1.24.0.
* Update build to Go 1.25.
* Update golangci-lint.
* Fixup linting issues.

### DIFF
--- a/pkg/probe/firewall_load_balance.go
+++ b/pkg/probe/firewall_load_balance.go
@@ -133,11 +133,12 @@ func probeFirewallLoadBalance(c http.FortiHTTP, meta *TargetMetadata) ([]prometh
 				}
 
 				realServerRTTValue := math.NaN()
-				if "<1" == realServer.RTT {
+				switch realServer.RTT {
+				case "<1":
 					realServerRTTValue = 0.001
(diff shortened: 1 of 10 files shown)


pkg/probe/firewall_load_balance.go
```diff
@@ -133,11 +133,13 @@ func probeFirewallLoadBalance(c http.FortiHTTP, meta *TargetMetadata) ([]prometh
...
 				realServerRTTValue := math.NaN()
-				if "<1" == realServer.RTT {
+				switch realServer.RTT {
+				case "<1":
 					realServerRTTValue = 0.001
-				} else if "" == realServer.RTT {
+				case "":
 					// NaN
-				} else {
+					fallthrough
```
sysadmind: `fallthrough` changes this behavior doesn't it? Before the default case code would only execute when RTT was NOT "<1" or ""

sysadmind (approved):

pkg/probe/firewall_load_balance.go
SuperQ: I think you're correct, not sure what I was thinking with this.

PR #381, opened 2026-03-20
fix: guard against empty firewall policy results to prevent panic

Aprazor:
If the FortiOS API returns an empty array (no VDOMs or permission
error), accessing ps4[0] panics with index out of range.

### DIFF
--- a/pkg/probe/firewall_policy.go
+++ b/pkg/probe/firewall_policy.go
@@ -80,6 +80,11 @@ func probeFirewallPolicies(c fortigatehttpclient.FortiHTTP, _ *TargetMetadata) (
 		return nil, false
 	}
 
+	if len(ps4) == 0 {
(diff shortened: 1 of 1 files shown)


bastischubert:
Thanks for spotting this! The panic is real, but we'd prefer to fix the root cause: `probeFirewallPolicies` discards `*TargetMetadata` (the `_` parameter) and re-parses the version from `ps4[0]` — which is what forces the index access in the first place.

Other probes (`bgp_neighbors.go`, `managed_switch.go`, etc.) use `meta.VersionMajor`/`VersionMinor` directly. If this probe does the same, `ps4[0]` is never accessed for version info, the panic becomes impossible by design, and the guard isn't needed.

Would you be open to updating the PR to use `meta` instead? Happy to help if needed.

bastischubert:
@Aprazor - does that sound ok for you to go the `meta` route instead?

PR #384, opened 2026-03-20
fix: use r.Selected instead of r.Reachable for NTP selected label

Aprazor:
All 7 MustNewConstMetric calls passed r.Reachable for both the
reachable and selected label positions. The selected label always
showed the reachable value, making it meaningless.


bastischubert:
sorry for reviewing late, the fix looks good to me, but the tests also need updates now - can change them as well so we can get this one merged?

PR #396, opened 2026-04-13
feat(probe): add remote certificate metrics via with_remote parameter

luispcoutinho:
Add with_remote=1 query parameter to the available-certificates API calls in probeSystemAvailableCertificates, enabling the endpoint to return remote certificates (type=remote-cer) alongside local ones.

Remote certificates were previously invisible to the exporter despite being visible in the FortiGate GUI, as the API omits them by default.


bastischubert (approved):
looks like a useful enhancement

Issue #415, opened 2026-06-16
Add SSID label on "fortigate_wifi_client_info"


Semitour:
Hi there,

Great exporter. I use it on 10+ Fortigate to get informations about Wifi connectivity from multiple locations. 
It would be great to have the SSID label on the fortigate_wifi_client_info metric, so we can manage usefull informations about it. In our environment, we have 40+ SSID and thousands of wifi clients connected on approx. 70 AP's.

I'm not a dev but this looks pretty easy to add in the pkg/probe/wifi_clients.go file, since the information is available in the Fortigate API.

Can someone manage to add it in the next release?
This would be great :)

Best regards

lazyb0nes:
hi @Semitour 
Can you provide testdata for us?, if you can we can probably fix that pretty easily but we need data to test with.

Semitour:
Hey @lazyb0nes ,

Please find some anonymized test data below. If you need more, i can provide it.


FortiOS: 7.4.12
Endpoint: /api/v2/monitor/wifi/client?vdom=*&start=0&count=1000

Sanitized sample:
```json
[
  {
    "http_method": "GET",
    "results": [
      {
        "ip": "198.51.100.1",
        "ip6": [
          "::"
        ],
        "wtp_name": "ap-0001",
        "wtp_id": "REDACTED",
        "wtp_radio": 2,
        "wtp_ip": "198.51.100.101",
        "wtp_control_ip": "198.51.100.101",
        "wtp_control_local_ip": "198.51.100.101",
        "vap_name": "ssid-0001",
        "ssid": "ssid-0001",
        "mac": "02:00:00:00:00:01",
        "11k_capable": false,
        "11v_capable": false,
        "11r_capable": false,
        "sta_maxrate": 286800,
        "sta_rxrate": 77400,
        "sta_rxrate_mcs": 6,
        "sta_rxrate_score": 26,
        "sta_txrate": 103200,
        "sta_txrate_mcs": 8,
        "sta_txrate_score": 35,
        "authentication": "pass",
        "uses_captive_portal": false,
        "captive_portal_authenticated": 0,
        "manufacturer": "ShenZhen",
        "data_rate_bps": 180600000,
        "data_rxrate_bps": 77400000,
        "data_txrate_bps": 103200000,
        "snr": 23,
        "idle_time": 3,
        "association_time": 1781872182,
        "bandwidth_tx": 492,
        "bandwidth_rx": 389,
        "lan_authenticated": false,
[...]

PR #438, opened 2026-09-06
feat(wifi): add ssid label to fortigate_wifi_client_info 

bastischubert:
see #415


Semitour:
Hello, Thank you for the update. 
Do you know when the next version will be released so we can implement this in our dashboards?

